PR #18, opened 2026-03-12
feat(admin): superadmin global SMTP settings for auth mail (magic links)

ALI-RUBASS:
## What
- add nested Global Settings tabs for superadmin: `Login-Funktion` and `SMTP & System-E-Mails`
- move superadmin SMTP UI to a dedicated global form (no tenant SMTP component reuse)
- wire global SMTP values to app settings context keys (`globalSmtp*`, `globalFeatureSystemNotificationEmailsEnabled`)
- add DE translation keys for global SMTP tab/labels

## Why
- tenant SMTP and superadmin/global SMTP must be separate
- global SMTP is needed for platform-level auth email flows (especially **magic links**)
- this keeps auth mail transport independent from tenant chat email settings

## Impact
- no change to tenant SMTP pages
- superadmin now has a dedicated global SMTP tab under `/admin/global-settings/smtp`

## Validation
- admin build passes and page is deployed in dev
- navigation and tab routing verified for superadmin global settings

## Notes
- backend/settingsadmin wiring for saving global SMTP values is handled in the related service work and deployed alongside this UI direction.


src/pages/GlobalSettings/index.tsx
```diff
@@ -17,28 +44,93 @@ export const GlobalSettingsPage = () => {
...
+    const initialValues = useMemo(
+        () => ({
+            globalFeatureSystemNotificationEmailsEnabled:
+                settings.globalFeatureSystemNotificationEmailsEnabled ?? false,
+            globalSmtpEnabled: settings.globalSmtpEnabled ?? false,
+            globalSmtpHost: settings.globalSmtpHost ?? '',
+            globalSmtpPort: settings.globalSmtpPort ?? '587',
+            globalSmtpSecure: settings.globalSmtpSecure ?? false,
+            globalSmtpUsername: settings.globalSmtpUsername ?? '',
+            globalSmtpPassword: settings.globalSmtpPassword ?? '',
```
shazia-k: Issue: Global SMTP password handling is unsafe or lossy. The new form reads globalSmtpPassword from app config context, but that context is loaded from unauthenticated server settings. If the backend exposes the password there, the secret becomes public. If it does not, the form initializes it as empty and a normal save can overwrite the stored password with "".
Suggestion: Load sensitive SMTP values from an authenticated admin-only endpoint, or treat the password as write-only and omit it from PATCH unless the user explicitly changes it.

src/pages/GlobalSettings/index.tsx
```diff
@@ -17,28 +44,93 @@ export const GlobalSettingsPage = () => {
...
+                        <FormSwitchField
+                            labelKey="globalSettings.smtp.smtpToggle.title"
+                            name={['globalSmtpEnabled']}
+                            inline
+                            disableLabels
+                        />
+                    </div>
+
+                    <FormInputField labelKey="globalSettings.smtp.host" name={['globalSmtpHost']} />
+                    <FormInputField labelKey="globalSettings.smtp.port" name={['globalSmtpPort']} />
```
shazia-k: Issue: Global SMTP port uses a plain text input instead of numeric validation. The tenant SMTP editor already uses a bounded numeric field, but the new global form accepts arbitrary strings.
Suggestion: Use FormInputNumberField with sane bounds, matching the tenant SMTP implementation.

shazia-k (commented):
@ALI-RUBASS suggested few changes in code. Please consider them before approval.

ALI-RUBASS:
Addressed the requested review items with minimal scoped changes in ORISO-Admin only:\n\n1) **Password safety/write-only behavior**\n-  is no longer initialized from settings context.\n- Save flow now omits  from PATCH unless user explicitly enters a value.\n- This prevents accidental empty-string overwrite and avoids exposing persisted secret via normal settings load path.\n\n2) **SMTP port validation**\n- Replaced plain text port input with  and bounds () to match tenant-side behavior.\n- Port is normalized to string before PATCH to keep backend contract stable.\n\nCommit:

ALI-RUBASS:
Addressed the requested review items with minimal scoped changes in ORISO-Admin only:

1) Password safety/write-only behavior
- `globalSmtpPassword` is no longer initialized from settings context.
- Save flow now omits `globalSmtpPassword` from PATCH unless user explicitly enters a value.
- This prevents accidental empty-string overwrite and avoids exposing persisted secret via normal settings load path.

2) SMTP port validation
- Replaced plain text port input with `FormInputNumberField` and bounds (`1..65535`) to match tenant-side behavior.
- Port is normalized to string before PATCH to keep backend contract stable.

Commit: `3a32f94`

shazia-k (approved):
Looks good!

PR #20, opened 2026-03-27
feat(security-06): add superadmin inactivity audit log page

ALI-RUBASS:
## Summary
- add a superadmin-only admin route/sidebar page to view Security-06 inactivity audit logs
- implement table pagination and filtering by `accountRole` and `accountId`
- wire frontend API integration and localization for the new log page

## Test plan
- [x] build `ORISO-Admin` successfully
- [x] deploy admin image and verify rollout in k8s
- [x] open `/admin/logs/inactive-accounts` as superadmin and verify page renders
- [x] verify non-superadmin users do not see the sidebar entry

Made with [Cursor](https://cursor.com)


src/hooks/useInactiveAccountAuditLogsData.ts
```diff
@@ -0,0 +1,47 @@
+import { QueryOptions, useQuery, UseQueryOptions } from 'react-query';
```
shazia-k: Issue: QueryOptions is imported but never used. On line 45 the options are cast as QueryOptions which is a different type from UseQueryOptions (the one expected by useQuery). This cast may silently drop type checks. Remove the unused QueryOptions import and the as QueryOptions cast — just pass { ...options, retry: false, refetchOnWindowFocus: false } directly since options already extends UseQueryOptions.

src/hooks/useInactiveAccountAuditLogsData.ts
```diff
@@ -0,0 +1,47 @@
...
+    });
+
+    if (accountRole) {
+        params.set('accountRole', accountRole);
+    }
+    if (accountId?.trim()) {
+        params.set('accountId', accountId.trim());
+    }
+
+    return useQuery(
```
shazia-k: The useQuery call has no generic type parameter, so the return type is inferred as any from fetchData(). This means data in the page component has no type safety. Add the generic: useQuery(...) so callers get proper typing on data.data, data.total, etc.

src/pages/Logs/InactiveAccountAuditLogs/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+                        setAccountRole(value);
+                        setPage(1);
+                    }}
+                    options={[
+                        { value: 'ASKER', label: 'ASKER' },
+                        { value: 'CONSULTANT', label: 'CONSULTANT' },
+                        { value: 'ADMIN', label: 'ADMIN' },
+                    ]}
+                />
+                <Input
```
shazia-k, line 118: The account ID search input is only applied on Enter or clicking Search. The hook does accountId?.trim() before sending to the API, but accountIdFilter retains the untrimmed value locally. Consider trimming when setting the filter (setAccountIdFilter(accountIdInput.trim())) so displayed state and queried value always match.

src/pages/Logs/InactiveAccountAuditLogs/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+                loading={isLoading}
+                columns={columns}
+                dataSource={data?.data ?? []}
+                scroll={{ x: 1600 }}
+                pagination={{
+                    current: page,
+                    pageSize: perPage,
+                    total: data?.total ?? 0,
+                    showSizeChanger: true,
+                    onChange: (nextPage, nextPageSize) => {
```
shazia-k, line 160: Pagination onChange: when nextPageSize changes, setPage(nextPage) is called first, then setPage(1) overwrites it. The first call is wasted. Simplify: only call setPage(nextPage) when page size hasn't changed; otherwise call setPage(1) and setPerPage(nextPageSize).

src/components/Layout/NavIcon.tsx
```diff
@@ -45,6 +45,7 @@ const Icon = ({ path, hover }: { path: string; hover: boolean }) => {
         case routePathNames.statistic:
             return currentlySelected ? <StatisticsActiveIcon /> : <StatisticsInactiveIcon />;
         case routePathNames.logs:
+        case routePathNames.inactiveAccountAuditLogs:
```
shazia-k: The inactiveAccountAuditLogs case falls through to the same Statistics icon as logs. The audit logs page uses the statistics icon rather than a security/audit-specific icon. If not intentional, consider using a distinct icon (e.g. AuditOutlined or SafetyOutlined from Ant Design) so superadmins can visually distinguish audit logs from statistics/supervisor logs.

src/pages/Logs/InactiveAccountAuditLogs/index.tsx
shazia-k: fixed

shazia-k (approved):
@ALI-RUBASS I committed the fixes and PR approves

PR #251, opened 2026-07-07
test(admin): cover the antd M3 theme bridge and Figma case-handover controls

kiodreambau:
## Summary

Follow-up unit-test coverage for the freshly merged #248 (Case Handover card matching Figma, new `M3Checkbox`) and #250 (antd M3 token bridge). No production code changes — tests only.

- **New `src/theme/antdM3Theme.test.ts`** (#250 landed with no unit test):
  - `buildAdminAntdTheme()` derives every antd colour token from the `computeOrisoPalette` M3 token set — the single source of truth — instead of the legacy hard-coded blue (`#273270`)
  - tenant colour seeds propagate into `colorPrimary`/`colorLink`
  - the `inverted` scheme selects `antdTheme.darkAlgorithm` and the inverted palette
  - fallback path: when the palette omits a token, the baked-in M3 defaults apply
  - M3 shape (`8/16/4` radii), Inter/Roboto typography, and the 40px control height (WCAG 2.2 target-size headroom) hold for both schemes
- **Extended `src/components/M3Checkbox/index.test.tsx`**: unchecking a checked box emits `onChange(false)`; keyboard operability (Tab focuses, Space and Enter toggle); disabled removes it from the tab order. Existing tests already covered role/checked exposure, click toggling and the disabled no-op.
- **Extended `CaseHandoverCard.test.tsx`**: the two new Figma "enforce" feature-row checkboxes render disabled + unchecked, sit inside keyboard-reachable tooltip wrappers (`tabindex="0"`), and clicking them never persists anything. The pre-existing tests (master toggle, consent toggles, placeholder tab, error/rollback paths) still pass unchanged against the new markup.

Deliberately **no CardDeck tests** — #249 ships its own and would conflict.

## Test plan

- [x] `npx vitest run` — 100 files / 471 tests green (18 in the touched files)
- [x] `npx tsc --noEmit` — clean
- [x] `npx prettier --check` on the changed files — clean
- Note: `npm run lint:js` is argument-less `eslint` (v7), a no-op; running eslint on any `*.test.tsx` file (pre-existing ones included) fails with a parsing error because `tsconfig.json` excludes test files from the type-aware parser project — a pre-existing repo-wide limitation, not introduced here.

### DIFF
--- a/src/components/M3Checkbox/index.test.tsx
+++ b/src/components/M3Checkbox/index.test.tsx
@@ -18,6 +18,37 @@ describe('M3Checkbox', () => {
         expect(onChange).toHaveBeenCalledWith(true);
     });
 
+    it('unchecks a checked box via onChange(false)', async () => {
+        const onChange = vi.fn();
+        const user = userEvent.setup();
+        render(<M3Checkbox checked label="Activated" onChange={onChange} />);
+        await user.click(screen.getByRole('checkbox', { name: 'Activated' }));
+        expect(onChange).toHaveBeenCalledWith(false);
+    });
+
+    it('is keyboard-operable: Tab focuses it, Space and Enter toggle', async () => {
+        const onChange = vi.fn();
+        const user = userEvent.setup();
+        render(<M3Checkbox checked={false} label="Activated" onChange={onChange} />);
+
+        await user.tab();
+        expect(screen.getByRole('checkbox', { name: 'Activated' })).toHaveFocus();
+
+        await user.keyboard(' ');
+        expect(onChange).toHaveBeenNthCalledWith(1, true);
+
+        await user.keyboard('{Enter}');
+        expect(onChange).toHaveBeenNthCalledWith(2, true);
+    });
+
+    it('is removed from the tab order when disabled', async () => {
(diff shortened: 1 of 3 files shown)


Kadyan25 (approved):
Test-only PR, cross-checked all three additions against real production code rather than trusting the diff:

- M3Checkbox is a native <button role="checkbox">, so the new keyboard-operability test (Tab focus, Space/Enter toggle) and the disabled-removes-from-tab-order test are just confirming real native button behavior plus the component's own onClick handler.
- CaseHandoverCard's new "Figma enforce checkbox" test: verified every assertion (disabled, aria-checked=false, wrapping span tabIndex=0, exact accessible-name format) against CaseHandoverCardView.tsx line by line - every "coming soon" checkbox is wrapped in <Tooltip><span tabIndex={0}><M3Checkbox disabled .../></span></Tooltip>, matches precisely.
- antdM3Theme.test.ts (new file): confirmed every fallback color/radius/font/control-height constant asserted matches the real t(name, fallback) defaults and constants in antdM3Theme.ts exactly, including dark/light algorithm branching.

No fabricated or mismatched assertions. Approving.

PR #252, opened 2026-07-07
feat(storybook): enforce WCAG 2.2 AA via @storybook/addon-a11y

kiodreambau:
## Why
The admin Storybook had no accessibility tooling: components could regress on WCAG without anyone noticing. Decision from 2026-07-07: a11y checks are mandatory for every story, targeting WCAG 2.2 AA.

## What
- Add `@storybook/addon-a11y` (axe-core) to `.storybook/main.ts`
- Configure it globally in `.storybook/preview.tsx`: rule set `wcag2a + wcag2aa + wcag21a/aa + wcag22aa`, `test: 'error'` so violations surface as errors in the Accessibility panel (and in `storybook test` once the vitest addon lands)

## Verification
- Storybook boots on current dev; the Accessibility tab scans every story
- Sample runs: `Molecules/Legal/LegalVersionViewer` → 0 violations / 18 passes; `Organisms/M3 Rich Text Editor` → the gate immediately caught a real **Serious 'ARIA input field name'** violation (TipTap contenteditable has no accessible name) — follow-up fix is spun off separately, deliberately not part of this infra PR
- `prettier --check` and `tsc --noEmit --project tsconfig.storybook.json` clean

## Note
The stray `yarn.lock` delta is the matching `addon-a11y` entry so both lockfiles stay consistent.

Closes #265


Kadyan25 (approved):
Pure dev-tooling addition: adds @storybook/addon-a11y, wires it into .storybook/main.ts, configures preview.tsx to run axe-core against WCAG 2.0a/2.0aa/2.1a/2.1aa/2.2aa tags with test: 'error'.

Checked whether test: 'error' risks silently blocking CI on pre-existing violations - it doesn't: no CI workflow runs Storybook at all, and there isn't even a "storybook" script in package.json yet. This only affects the local Storybook dev experience, zero blast radius on CI or the running app.

CI green on the actual build job. Approving.

shanzaeimran (approved):

PR #275, opened 2026-07-08
test(agency-list): lock in graceful degradation on a failed agencies load (#240)

kiodreambau:
## Context
#240 reports the agency list at `/admin/agency` spinning forever. The page already reads `isError`/`isLoading` from `useAgenciesData` and shows an error `emptyText`, and the query uses `retry: false`. The one way the spinner could hang indefinitely was a request that never settles — now prevented by fetchData's 30s default timeout (#143): after at most 30s the query resolves to `isError`, `isLoading` becomes false, and the error message shows.

## What
- Regression test: a failed agencies load shows the error message and is **not** in the loading state.
- Test: a successful empty load shows the neutral empty text, not the error.
- The `ResizeTable` test mock now renders `loading` + `locale.emptyText` so these states are assertable (previously it only rendered rows).

Full suite: 496/496; eslint + prettier clean.

## Scope
This nails the **frontend symptom** (no more infinite spinner — it degrades to a visible error). If the list is still empty on `admin.oriso.org`, the **root cause is backend** (agencies API returning 500 / not responding — cf. the UserService agency-admin self-call config drift) and should be reproduced against the live service separately. Leaving #240 open for that root-cause confirmation rather than closing on the frontend guard alone.

Part of #240

### DIFF
--- a/src/pages/Agency/List/index.test.tsx
+++ b/src/pages/Agency/List/index.test.tsx
@@ -114,7 +114,7 @@ vi.mock('../../../components/ResizableTable', () => ({
         return (
             <div>
                 {loading && <div data-testid="table-loading" />}
-                {dataSource.length === 0 && <div data-testid="table-empty">{locale?.emptyText}</div>}
+                {!loading && dataSource.length === 0 && <div data-testid="table-empty">{locale?.emptyText}</div>}
                 {dataSource.map((record: any) => (
                     <div key={record.id}>
                         <div data-testid={`topics-${record.id}`}>{topicColumn.render(record.topics, record)}</div>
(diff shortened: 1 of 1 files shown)


radaljj (approved):
LGTM overall — this is a focused regression test for the agency list error state and matches the current pre-dev behavior: when the query settles with isError=true, loading is false and the table shows the default error empty text instead of staying on a spinner.

Non-blocking: the ResizeTable mock could be made slightly closer to the real table by rendering emptyText only when !loading && dataSource.length === 0. Current tests are fine, but that would avoid future tests accidentally asserting a loading+empty state that the real UI should not show.

PR #278, opened 2026-07-08
test(user-list): lock in the sort fallback that prevents endless loading (#77)

kiodreambau:
## Context
#77: sorting the Counsellor / Counsellor-Admin lists spun forever when the backend rejected the sort field. The fix is already in place (`fetchUserSearchWithSortFallback` + `retry: false` + fetchData's 30s timeout from #143) but had **no test**.

## What
Adds `fetchUserSearchWithSortFallback.test.ts` covering the anti-hang contract:
- success returns the list with the mapped `field`/`order`
- a rejected preferred field (`UPDATE_DATE`) falls back to the API-safe `FIRSTNAME`/`ASC`
- **double failure returns an empty list — never hangs or throws** (the #77 core)
- no pointless second request when the primary already used the safe sort
- `rethrowOnFailure` surfaces the error when asked
- a provided `normalizeSortField` guard is applied

Full suite: 500/500; eslint + prettier clean.

Closes #77

### DIFF
--- a/src/utils/fetchUserSearchWithSortFallback.test.ts
+++ b/src/utils/fetchUserSearchWithSortFallback.test.ts
@@ -1,0 +1,105 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const fetchDataMock = vi.fn();
+vi.mock('../api/fetchData', () => ({
+    fetchData: (...args: unknown[]) => fetchDataMock(...args),
+    FETCH_METHODS: { GET: 'GET' },
+}));
+// removeEmbedded just unwraps the HAL envelope; pass the list through.
+vi.mock('./removeEmbedded', () => ({ default: (r: unknown) => r }));
+
+// eslint-disable-next-line import/first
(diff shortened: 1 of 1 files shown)


Kadyan25 (approved):
Approved. Verified this test file (`fetchUserSearchWithSortFallback.test.ts`) against the real production implementation at `src/utils/fetchUserSearchWithSortFallback.ts` — all 6 cases match actual behavior, not just asserting whatever the code currently does:

- Default sort passthrough (mapped field/order in the request URL)
- Fallback to the safe sort (`FIRSTNAME`/`ASC`) when the requested field is rejected
- Empty-list return (never throws/hangs) when both primary and fallback fail
- No redundant retry when the primary request already used the safe sort
- `rethrowOnFailure` correctly propagates instead of swallowing
- `normalizeSortField` override respected ahead of the default field resolution

No production code changed, no regressions introduced. Locks in real coverage for the #77 sort-hang fix.

Issue #229, opened 2026-07-02
[UI-MOD-01B] Admin Cypress upgrade validation


Storypapst:
Affected repos: OpenResilienceInitiative/ORISO-Admin

## Why
The Admin modernization branch upgrades Cypress alongside Vite and MUI. The upgrade must be locally verified before the PR is considered review-ready.

## What to build
- Pull the freshest Admin Vite/Cypress/MUI 9 branch locally.
- Compare it against origin/dev and identify merge conflicts or dependency drift.
- Run the local Cypress CLI path and repair bounded upgrade issues.

## Acceptance
- [ ] Admin modernization branch is fetched and compared against dev.
- [ ] Cypress version and scripts are documented.
- [ ] Local Cypress run succeeds or concrete blockers are documented with follow-up steps.
- [ ] Part of epic OpenResilienceInitiative/ORISO-Frontend#353

PR #280, opened 2026-07-08
test(cypress): fix stale login spec selectors after MUI login-form migration (#229)

kiodreambau:
## Context
Validating the Cypress upgrade (#229). Cypress 15.18 verified + runs, but the one login smoke spec still targeted the old antd field ids (`#basic_username` / `#basic_password`) which the MUI-migrated `LoginForm` no longer emits (MUI/antd produce auto-generated ids, no `name`). The spec was red against current code.

## What
- Switch to stable **semantic selectors**: `input[autocomplete="username"]` / `input[autocomplete="current-password"]` (these attributes are explicitly set on the fields and survive the id churn).
- Password-visibility toggle already uses `data-testid`, kept as-is.

## Verification (real E2E via Docker)
- Built `build/` + `docker build` → `docker run oriso-admin -p 9000:80` (nginx serves `/admin/login`, HTTP 200)
- `cypress run` against it: **3 passing, 2 pending** (the credential-gated authenticated suite skips) — was 3 failing before.
- Cypress binary `cypress verify` ✓ (15.18.0).

Part of #229

### DIFF
--- a/cypress/integration/login.spec.js
+++ b/cypress/integration/login.spec.js
@@ -5,29 +5,29 @@ describe('login page', () => {
 
     it('displays username and password fields', () => {
         cy.get('form').should('be.visible');
-        cy.get('#basic_username').should('be.visible');
-        cy.get('#basic_password').should('be.visible');
+        cy.get('input[autocomplete="username"]').should('be.visible');
+        cy.get('input[autocomplete="current-password"]').should('be.visible');
     });
 
     it('keeps submit disabled until username and password are filled', () => {
         cy.get('button[type="submit"]').should('be.disabled');
-        cy.get('#basic_username').type('test-user');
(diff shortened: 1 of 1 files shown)


Kadyan25 (approved):
Approved. Live-verified against the running dev server:

- `input[autocomplete="username"]` and `input[autocomplete="current-password"]` both exist and are the actual rendered MUI fields
- The old selectors (`#basic_username`, `#basic_password`) are confirmed gone — genuinely would fail today, confirming this fix is necessary
- Clicked `[data-testid="password-visibility-toggle"]` — password field type flips `password` → `text`, toggle's `aria-pressed` flips to `true`, matching the spec's third test case exactly

All three updated Cypress assertions hold against the real app.

Issue #268, opened 2026-07-08
M3 editor: wire the version select — load and display saved versions


kiodreambau:
## Current state
- Published legal versions ARE saved server-side and are selectable in the read-only `LegalVersionViewer` (e.g. the DPA card loads them via `useDpaVersions`).
- The version dropdown inside `M3RichTextEditor` ("Latest Version") is still a stub: 'latest'/'published' do nothing and 'version history' is a disabled entry.

## Goal
Make the editor's version select functional:
- Props for a version list (id, label, content) + `onVersionSelect`
- Selecting a version shows its content (read-only) with a clear way back to the current draft
- Containers (DPA / imprint / privacy cards) feed their existing version data in
- Decide: is restoring an old version as a new draft in scope?

Stories for every state (no versions, several versions, viewing an old version) + tests per the module-state convention.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Agreed semantics (2026-07-08, Frank)
- **Restore = copy**: selecting an old version never mutates it — restoring duplicates its content into the editor as a new draft; publishing appends a NEW version. The version chain stays append-only/immutable (backend already works this way: `TenantDpaVersionEntity.activationDate` identifies each published snapshot).
- **Dates**: the saved/published date exists (`activationDate`) and is already shown in the version select. The active range (from–until) is derivable — *until* = the next version's `activationDate` — and should be displayed in the select/viewer; explicit storage only if legal compliance requires it.
- **Dynamic status description**: the card description must react to the document status (e.g. never published / published / editing a restored version) in DE+EN — Figma node to be provided by Frank.

Storypapst:
Partially delivered: the referenced PR(s) (ORISO-Admin#282 ORISO-Admin#346 ORISO-Admin#343) are merged, but acceptance criteria remain open. Moving to **QA internal Dev** for an internal re-check of the remaining scope.

PR #255, opened 2026-07-08
feat(m3-editor): standard anchor navigation (heading anchors + horizontal jump menu + link-to-anchor bubble)

kiodreambau:
> **Stacked on #254** (which is stacked on #253) — merge those first; until then this diff includes their commits.

## Why
Anchor navigation existed only in the form-bound `TiptapEditor` (legal cards, #234/#246). It is now a **standard feature of the `M3RichTextEditor`** so every long text gets jump marks and a horizontal section menu — as requested, without new dependencies.

## Approach — deliberate deviation from the proposed extensions
`@tiptap/extension-unique-id` and `@tiptap/extension-table-of-contents` were evaluated and rejected for documented reasons (see the note in `headingAnchors.ts`): UniqueID instantly re-assigns ids that a user removed, and the ToC extension lists headings regardless of anchor state. The repo's own `HeadingAnchors` extension already implements the exact semantics (readable slugs, persistent removal, duplicate suffixing) — reused instead of duplicated. The BubbleMenu + Link part follows the proposal 1:1.

## What
- **New `useHeadingAnchorNav` hook** — the anchor behavior extracted from `TiptapEditor` (anchor list sync, legacy-content stamping, read-only scroll spy, jump/remove, #link click interception). `TiptapEditor` now consumes the hook (net deletion there); both editors share one implementation.
- **`M3RichTextEditor`**: new `enableAnchors` prop (default **true**); `HeadingAnchors` extension; `AnchorChips` row between toolbar and editor — **single line, hidden horizontal scrollbar** (same pattern as the toolbar/segmented tabs); BubbleMenu with an anchor `Select` on text selection (edit mode); scroll spy + link interception (read-only). Skipped when an `editorSlot` owns the content.
- `AnchorChips` accepts a `className` for the M3 chrome.

## Test coverage (all module states)
14 new tests in `M3RichTextEditor.anchors.test.tsx`:
- **Unit/Integration (edit)**: chip row renders per anchored heading (removable), chip click scrolls, removal strips the id + persists `data-anchor-removed`, legacy content gets slug ids, duplicate headings get `-2` suffixes
- **States**: empty content (no row), `enableAnchors={false}`, `editorSlot` (no internal row), read-only (no "x", checkmark on jump, no id invention), fullscreen dialog keeps the anchor row
- **Contract**: HTML round-trip — emitted HTML re-rendered yields identical anchors; a removed anchor stays removed (no re-assignment)
[...]

PR #282, opened 2026-07-09
feat(m3-editor): wire the version select — browse + restore saved versions (#268)

kiodreambau:
> **Stacked on #255 → #254 → #253** — merge those first; until then this diff includes their commits.

## Why
You asked for the editor's version to be selectable via the select field. The dropdown was a stub; this wires it to real saved versions.

## What (additive to `M3RichTextEditor`)
- New props `versions: {id,label,content}[]` + `onRestoreVersion(content)`.
- Version menu lists **current draft** + each saved version (by date label).
- Selecting an older version shows it **read-only** in the editor with a look-back banner.
- **Restore as new draft** = copies the content into the draft (append-only; the old version is never mutated — the restore=copy semantics you confirmed).
- **Back to the current version** returns to the editable draft.
- Read-only cards (agency view) can browse but never restore.

## Deliberately out of scope (needs your input)
- The **dynamic status-description** text (never-published / published / viewing-restored) — awaits the Figma node you mentioned.
- Whether this **replaces the separate `LegalVersionViewer`** under the card: left additive so the container decides — no existing UI removed.

## Verification
- TDD: 6 red→green tests (`M3RichTextEditor.versions.test.tsx`); full suite **499/499**; tsc + eslint + prettier clean.
- Stories: `WithVersionSelect`, `VersionSelectReadOnly`.

Part of #268


radaljj (approved):
## Verdict: Approve (with process notes)

The tip commit for #268 (version select) is clean, additive, and well-tested. Restore=copy semantics match the agreed model. CI green.

### What looks good
- Clear contract: `versions` + `onRestoreVersion(content)` — look-back is read-only; restore copies into the draft; history is never mutated.
- `editorEditable = !readOnly && !viewingVersion` correctly disables toolbar / publish while browsing an older version; draft `value` is preserved.
- Banner (`role=status`) + restore / back-to-draft controls; read-only cards can browse but never restore.
- TDD coverage in `M3RichTextEditor.versions.test.tsx` (menu, view, restore, back, empty, read-only).
- Stories: `WithVersionSelect`, `VersionSelectReadOnly`.
- i18n keys (de/en) look fine.

### Process / merge blockers (not code nits)
1. **Stacked on #253 → #254 → #255** — all still open. Do not merge this PR alone; merge the stack in order (or rebase this tip onto `pre-dev` after those land).
2. **Base is `dev`** — per current policy, retarget to `pre-dev` before merge.
3. **Not wired to product yet** — `DataProcessingAgreementCard` still uses separate `LegalVersionViewer` and does not pass `versions` / `onRestoreVersion` into `M3RichTextEditor`. Fine as an additive editor capability (as noted in the PR), but #268 is not end-to-end until a follow-up wires the container.

### Minor (non-blocking)
- While viewing an old version, `AnchorChips` still gets `editable={!readOnly}` instead of `editorEditable`, so remove (“x”) can still appear even though the editor is non-editable. Prefer `editable={editorEditable}`.
- If `versions` is updated and drops the currently selected id, `viewingVersion` becomes null but `viewingVersionId` can stay stale until the next menu click — consider resetting when the id is missing from `versions`.

### Out of scope (agreed)
- Dynamic status-description copy (Figma).
- Replacing `LegalVersionViewer` — left to the container.

Happy to re-check once the stack is on `pre-dev` and/or the DPA card wires these props.

radaljj (changes requested):
Please rebase the branch to pre-dev instead of dev

src/components/FormPluginEditor/M3RichTextEditor.tsx
```diff
@@ -420,20 +476,90 @@ export const M3RichTextEditor = ({
...
-            ) : (
-                <div className={styles.editorWrap}>
+            {aboveEditorSlot}
+
+            {anchorsEnabled && (
+                <AnchorChips
+                    className={styles.anchorNav}
+                    anchors={anchors}
+                    activeId={activeAnchorId}
+                    editable={!readOnly}
```
shanzaeimran: Confirming @radaljj's point, and it's more than cosmetic — it's a draft data-loss path. While viewing a saved version in an editable card (readOnly=false, viewingVersion set), the chips still render their remove-"x". Clicking it calls editor.commands.removeHeadingAnchor(id), which dispatches a transaction even though the editor is non-editable (TipTap `editable=false` only blocks user typing, not programmatic commands). That fires onUpdate → onChange(getHTML()), so the parent's onChange (e.g. setValue) is set to the version's content minus that anchor — the live draft gets silently overwritten with the old version.

Fix: editable={editorEditable}.

src/components/FormPluginEditor/M3RichTextEditor.tsx
```diff
@@ -420,20 +476,90 @@ export const M3RichTextEditor = ({
...
+
+            {editorSlot ?? (
+                <div className={styles.editorWrap} onClickCapture={handleContentClickCapture}>
                     <div className={styles.editor}>
                         <EditorContent editor={editor} />
                     </div>
                 </div>
             )}
 
+            {anchorsEnabled && !readOnly && (
```
shanzaeimran: Same root cause as the chips, second vector: the anchor-link bubble also gates on `!readOnly`, so it stays active while viewing a version. Setting a link mutates the doc the same way → onChange overwrites the draft with the version content. (This one depends on ProseMirror surfacing a non-empty selection on a non-editable view, so it's a bit less certain than the chips — but gating on editorEditable closes it regardless.)

Fix: {anchorsEnabled && editorEditable && ( … )}

Note editorEditable is already used correctly for the Toolbar and the
publish/save footer — these two anchor controls are the only spots still on
`!readOnly`.

src/components/FormPluginEditor/M3RichTextEditor.versions.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+        expect(screen.queryByText(/legal\.m3Editor\.viewingVersion/i)).not.toBeInTheDocument();
+    });
+
+    it('does not show restore/back controls when no versions are provided', async () => {
+        render(<M3RichTextEditor title="Impressum" value="<p>x</p>" enableAnchors={false} />);
+        expect(screen.queryByText(/legal\.m3Editor\.viewingVersion/i)).not.toBeInTheDocument();
+        expect(screen.queryByRole('button', { name: /legal\.m3Editor\.restoreVersion/i })).not.toBeInTheDocument();
+    });
+
+    it('never offers restore in read-only mode even if versions are passed', async () => {
```
shanzaeimran, line 125: Neither mutation vector above is covered — the version tests assert contenteditable=false and the banner, but never try removing an anchor (or linking via the bubble) while viewing a version. Suggest adding a test: view a version, click a chip's remove-x, and assert onChange is NOT called (and the draft value is unchanged).

PR #292, opened 2026-07-11
feat(permissions): enforce-on UI + toggle-revert root fix + null/cache fixes (P0+P2 of #291)

kiodreambau:
Admin foundation (P0) + toggle-revert root fix (P2) for the three-level permission cascade — see #291 / ADR-013.

**Problem:** No way to *force a feature on*; a toggle change could **jump back after save**; the card crashed without an app-config provider.

**What changed**
- fix (P2): `syncMasterTogglesToTenantAdminControls` master-gated encoded toggles (`<master>Enabled && isEnabled(field)`), so editing a sub-feature while its card master was off silently reset it to false on save — the confirmed "toggle springt zurück" bug. Now each toggle maps 1:1 to its own field (`TOGGLE_KEY_TO_FIELD`); master→children stays a UI-only concern.
- fix: `useTenantAdminControlsMutation` cached the request body instead of the server response.
- fix: `useAppConfigContext` null-context crash → degrades to cluster defaults.
- feat (P0): enforce-on logic `getEnforcedOnFields`/`applyEnforcedOnFields`; `PermissionsSettingsView` enforce mode (per-feature checkbox + footer, Figma `105-11334`); en/de i18n.
- feat: `PermissionsSettings` Storybook stories (5 states) linked to Figma `105-11334` & `105-11674` with an interaction test — the chat-type cards had no stories before.

**Verification:** vitest 511/511, `eslint --max-warnings=0`, `prettier --check`, `npm run build` — all green.

Closes P0 + P2 of #291.


src/components/Tenants/AppSettings/PermissionsSettings/permissionsSettingsUtils.ts
```diff
@@ -137,6 +137,62 @@ export const applyForcedOffFields = (
     return nextSettings;
 };
 
+/**
+ * Mirror of {@link getForcedOffFields} for the "enforce active state" direction: an upper role
+ * (platform → tenant → agency) can lock a feature *on* so lower roles cannot hide it. A truthy
+ * enforcement flag expands to every concrete feature field, which the lower level then renders
+ * as on-and-disabled.
+ */
+export const getEnforcedOnFields = (enforcedToggles?: PermissionToggleVisibility) => {
```
shanzaeimran, line 146: Encode/decode aren't inverse for group-master enforce checkboxes, which
reintroduces a "toggle springs back" bug on the enforce path.

- decode `getEnforcedOnFields` expands a key through PLATFORM_TOGGLE_FIELDS, and
  group keys expand to many fields — e.g. `anonymousChat` → 6 fields
  (featureAnonymousChatEnabled + video/audio/voice/threads/supervision-in-anonymous).
- encode `enforcedFieldsToToggles` collapses each field to one key 1:1 via
  FIELD_TO_TOGGLE_KEY.

The Live-chat / Group / Supervision card masters bind to group keys
(chatTypeCards.ts: featureAnonymousChatEnabled→anonymousChat,
featureGroupChatV2Enabled→groupChat, featureSupervisionEnabled→supervision), so:

1. Admin checks ONLY the Live-chat master enforce box → enforcedFields =
   {featureAnonymousChatEnabled}.
2. Save → enforcedFieldsToToggles → {anonymousChat: true}.
3. Reload → getEnforcedOnFields({anonymousChat:true}) → all 6 anonymous fields →
   every anonymous sub-feature enforce box now shows checked, and
   applyPermissionConstraintsToSettings locks video/audio/voice/threads/
   supervision-in-anonymous ON+disabled for every lower role — none of which the
   admin selected.
4. Worse: with the master enforced, unchecking a single sub reverts on reload,
   because the master field re-encodes to the group key which re-expands to all
   sub-fields.

Leaf-key sub-toggles round-trip cleanly; only group-master enforce boxes
misbehave. Suggest making encode mirror decode's grouping, or giving the enforce
path a strictly 1:1 field map (like TOGGLE_KEY_TO_FIELD) instead of reusing the
group-expanding PLATFORM_TOGGLE_FIELDS for decode.

src/components/Tenants/AppSettings/PermissionsSettings/permissionsSettingsUtils.test.ts
shanzaeimran, line 1: The enforce tests only cover leaf keys (videoCallsOneOnOneChats, audioCalls),
which is why the group-master round-trip bug above slips through green. Please add
a test that enforces a card master (e.g. featureAnonymousChatEnabled), runs
enforcedFieldsToToggles → getEnforcedOnFields, and asserts the set is unchanged
(no amplification to the other anonymous sub-fields).

src/hooks/useTenantAdminControlsMutation.hook.ts
```diff
@@ -21,7 +21,10 @@ export const useTenantAdminControlsMutation = ({
         mutationFn: updateTenantAdminControls,
         ...options,
         onSuccess: (responseData, updatedData, onMutateResult, context) => {
-            queryClient.setQueryData([TENANT_ADMIN_CONTROLS_KEY], updatedData);
+            // Cache the server's normalized response (backend fills every toggle via nullAsTrue and
```
shanzaeimran, line 24: Caching `responseData` is the right call for the nullAsTrue normalization, but it
can wipe the cache if the PUT ever resolves to an empty/undefined body: the old
code cached `updatedData` (always defined), so this is a behavior change. Consider
`setQueryData([TENANT_ADMIN_CONTROLS_KEY], responseData ?? updatedData)` to keep
the intent without blanking the controls until a refetch.

PR #299, opened 2026-07-13
fix(admin): stop tenantadmin prefetch on global-config settings page

shazia-k:
# Pull Request

**Click the `Preview` tab and select a PR template:**

- [Default / Generic](?expand=1&template=default.md)
- [Frontend](?expand=1&template=frontend.md)
- [Backend](?expand=1&template=backend.md)
- [Documentation](?expand=1&template=docs.md)
- [Bug Fix](?expand=1&template=bugfix.md)
- [DevOps / Infrastructure](?expand=1&template=devops.md)
- [Architecture / Refactor](?expand=1&template=architecture.md)
- [Hotfix](?expand=1&template=hotfix.md)

---

**⚠️ Please select a template above by clicking on one of the links.**

### DIFF
--- a/src/hooks/useSingleTenantData.test.tsx
+++ b/src/hooks/useSingleTenantData.test.tsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { renderHook, waitFor } from '@testing-library/react';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+
+vi.mock('../api/tenant/getSingleTenantData', () => ({
+    getSingleTenantData: vi.fn(),
+}));
+
+import { useSingleTenantData } from './useSingleTenantData';
+import { getSingleTenantData } from '../api/tenant/getSingleTenantData';
+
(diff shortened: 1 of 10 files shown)


Shirloin (commented):
The `prefetchTenantAdminData: false` flag doesn't actually stop the prefetch.

The mutation now passes `enabled: false` into `useSingleTenantData`, but that hook overrides it:

    // src/hooks/useSingleTenantData.ts:15
    enabled: (id && id !== 'add') || options.enabled,

On the global-settings page `id` is `"1"`, so `enabled` resolves to `true` regardless of the `false` we pass, and `GET /service/tenantadmin/1` still fires (and still 500s). Saves only succeed because they fall back to `seedTenantAdminData` — the failing request the PR is meant to remove is still sent on every render.

Fix belongs in `useSingleTenantData` so an explicit `enabled` is respected, e.g.:

    enabled: id != null && id !== 'add' && (options.enabled ?? true),

Please also add a hook-level test asserting no admin GET fires when `prefetchTenantAdminData: false`, since the current util tests don't cover this path.

PR #406, opened 2026-07-19
Fix: safe empty state for inaccessible/non-existent agencies in Admin

Storypapst:
## Summary
- `/admin/agency/:id` for a 403 (foreign agency) or a 404 (non-existent agency) used to render the normal edit form with empty/stale fields plus a generic "Something went wrong" toast.
- 403 and 404 are now surfaced identically (no toast, no distinguishable behavior) and render a dedicated "not available" empty state with a "Back to overview" action instead of the form.
- Other failures (network, 5xx) keep the existing generic toast behavior unchanged.
- No backend authorization changes; tenant separation is unchanged.

## What changed
- `fetchData.ts`: added an opt-in `FORBIDDEN_SILENT` response-handling flag so a caller can skip the global `/admin/access-denied` hard redirect on 403 and handle it inline. Additive only — every existing caller keeps the redirect exactly as before.
- `getAgencyById.ts`: 404 and 403 now both reject with a new `AgencyAccessError` (no toast); any other failure still shows the generic error toast.
- `Agency/Edit/index.tsx`: on `AgencyAccessError`, suppresses tabs/edit buttons/form and renders a `DashboardEmptyState` with a "Back to overview" button instead.
- Added `agency.edit.notFound.*` i18n keys (en/de).

## Test plan
- [x] `npx vitest run` (focused: `src/pages/Agency`, `src/api/agency/getAgencyById.test.ts`, `src/api/fetchData.test.ts`) — all green, including new 404/403/success/stale-data tests
- [x] `npx tsc --noEmit`
- [x] `npx eslint . --max-warnings=0` on changed files
- [x] `npm run build`
- [ ] Live browser screenshots — skipped (no Pre-Dev admin credentials available in this session)

## Summary by CodeRabbit

* **New Features**
  * Added a clear “agency not found or inaccessible” state with localized messaging and a return-to-overview action.
  * Prevented editing controls from appearing when access is unavailable.
  * Added silent handling for forbidden requests without automatic redirection.

* **Bug Fixes**
  * Improved handling of agency access errors and server failures.
  * Prevented stale agency data from appearing when switching between agencies.

* **Tests**
  * Added coverage for successful, denied, missing, and server-error scenarios.


Shirloin (changes requested):
**🚨 Blocker**
`useAgencyPostCodesData` always fires alongside the agency fetch 
and still uses old 403 handling — postcode 403 redirects to 
`/admin/access-denied`, defeating the fix.

Fix: gate postcode fetch behind successful agency load:
`enabled: Boolean(agencyData) && id !== 'add'`

**⚠️ Should fix**
`CATCH_ALL_SILENT` + local `message.error` can double-toast on 
timeout/network errors. Use `CATCH_ALL` instead and only map 
403/404 → `AgencyAccessError`.

- Page title duplicates "Back to overview" when inaccessible
- Consider reusing `FORBIDDEN` instead of new `FORBIDDEN_SILENT`
- Strengthen stale-data test with real React Query + shared client

**Missing tests**
- Postcode parallel fetch on inaccessible agency
- Timeout/network errors don't double-toast
- Tabs absent on inaccessible agency

Issue #572, opened 2026-07-28
[TEN-INV-U10] Global non-bypassable DPA blocker after login


kiodreambau:
**Priority:** Urgent *(Roadmap Priority field has no options yet — see parent)*

## Why
A manually created tenant admin with a password but no signed DPA currently gets an incomprehensible login/usage failure. A generic or silent login error is not acceptable.

## What to build
- An existing tenant admin with valid credentials can authenticate; with missing/outdated DPA a global screen blocker then locks the whole Admin app until signing.
- Only allowed: view/sign the DPA, retry, log out. Direct URLs and writing backend operations must not bypass the blocker.
- Clearly separate: invalid credentials · incomplete invite flow · successful login with missing DPA · technically inconsistent account. For invalid or not-yet-registered credentials show one combined privacy-preserving hint (check credentials, or finish the invite link first).
- Use current shared Storybook MUI components; verify mobile.

## Acceptance
- [ ] Tenant B (created with password, no DPA) logs in, sees the blocker, direct navigation stays locked, signing lifts the block permanently.
- [ ] No write operation succeeds while the blocker is active.

Part of OpenResilienceInitiative/ORISO-Admin#569

## Scroll behaviour (added 2026-07-28)

- [ ] The blocker content (full DPA text + form) is vertically scrollable inside the overlay at 390×844 while the app behind stays scroll-locked; sign/retry/logout are always reachable. Overlay must not use a fixed height with hidden overflow. Storybook story with an extra-long DPA text at the mobile viewport documents this.

## DPA anchor navigation (added 2026-07-28, Frank)

- [ ] Long DPA/AVV texts get in-document anchor navigation for fast skimming: a TOC generated from the legal content's section headings (sticky/side placement on desktop, compact dropdown or chip bar at 390×844) that smooth-scrolls to the target section INSIDE the scrollable text region; keyboard accessible (focus moves with the jump); shared implementation in the DPA form component so onboarding step and blocker both get it; Storybook story with a long multi-section text on desktop and mobile.

## DPA scope rule (Frank, 2026-07-29) — binding

[...]

Issue #569, opened 2026-07-28
[TEN-INV] Tenant admin onboarding chain broken: wrong invite link, false SENT status, unsafe ID allocation, missing DPA gate


kiodreambau:
**Affected repos:** ORISO-Admin · ORISO-UserService · ORISO-TenantService · ORISO-AgencyService · ORISO-ConsultingTypeService · ORISO-Frontend · ORISO-Helm · ORISO-E2E

**Priority:** Urgent · **Status:** In progress · **Owner:** @Storypapst (initial)

## Problem

Tenant (Träger) admin onboarding is broken as a **chain**, not as one isolated defect: ID assignment, invitation, email delivery, target routing, DPA (AVV) signing and login do not form a working flow on PreDev. The combination of a wrong link target, a false send status, possible ID collisions and an incomprehensible DPA lockout is **release-blocking**.

## Observed behaviour

- In **Tenant invites**, a platform admin can enter a person, pick an email template and trigger "Send directly". The UI then confirms the invitation was sent — but in a test with a real mailbox **no email arrived**. The system reports success without proven delivery (SMTP errors are swallowed, status shows `SENT`).
- The generated link from the link column routes to the **wrong app layer**: a tenant admin must complete invite + DPA on a **public Admin page**, but the link targets the App layer. The tenant is an organisation, not an app user; the login belongs to the tenant admin. The current redirect mixes these layers and makes the invite practically unusable.
- Platform admins can also create a tenant and a tenant admin **directly with a password**. That admin then holds credentials but no signed DPA and cannot understand why login or usage fails. There is no separation between: invalid credentials · an unfinished invite flow · a successful login with a missing DPA · a technically inconsistent account. Today the failure is generic/silent.
- The ID fields suggest tenant and agency IDs are freely selectable, without authoritatively showing whether an ID is **already assigned or reserved by an open invite** → data-integrity risk (collisions).
- **Evidence gap:** last week's E2E evidence documents *a* DPA flow, but does not prove today's failing end-to-end chain (tenant invite → real email → correct target route → account creation → DPA → login). This discrepancy between existing tests and real PreDev behaviour is explicitly part of this bug.

## User impact

Tenant onboarding is effectively impossible: invited tenant admins never receive a usable invitation, manually created tenant admins are locked out without explanation, and platform admins risk ID collisions that corrupt tenant/agency data.

### ID allocation contract (identical for tenant and agency IDs)

[...]

PR #588, opened 2026-07-29
feat(dpa): global non-bypassable DPA blocker after login

kiodreambau:
## Problem

A tenant admin who was created manually (not through the invite flow) could log in with valid credentials while having no DPA/AVV signed at all, or an outdated one. Instead of a comprehensible blocker they hit incoherent failures deeper in the app, and — worse — they could reach admin pages and mutating actions without a governing DPA. Deep links bypassed whatever partial guarding existed. This is the last gap of the broken onboarding chain in OpenResilienceInitiative/ORISO-Admin#569.

## What changed

- **`DpaBlockerGate` wraps the whole protected route tree** in `App.tsx`. For tenant-scoped admins the authoritative status (TenantService U9, `GET /tenantadmin/{id}/dpa/status`) is resolved *before any route renders*. A direct URL lands on the blocker, not on the page; while blocked, no admin UI — and therefore no mutating action — is reachable. The gate **fails closed**: unknown status, fetch error or an unresolved query never lets the app through. Backend write enforcement stays U9/U3 scope; this is the frontend lock.
- **`DpaBlocker` renders per status**: `UNSIGNED` / `OUTDATED` reuse the existing DPA form (published text via `/dpa/versions`, signer fields as in the onboarding step) and sign via `POST /tenantadmin/{id}/dpa/sign`; the returned `VALID` status is written into the query cache, lifting the block immediately and durably. `MISSING`, `INCONSISTENT` (technically inconsistent account) and `STATUS_UNAVAILABLE` are distinct retry/logout states. Retry and logout are always available.
- **Scroll behaviour** (#572 addendum): the overlay itself is the scroll container — no fixed height with hidden overflow — so the full DPA text plus form scroll at 390×844 while the body stays scroll-locked behind.
- **Login UX**: invalid credentials and a not-yet-registered invitee now get one combined, privacy-preserving inline hint (check credentials, or finish the invitation link first) — no user enumeration. Network, admin-only and tenant-access failures stay differentiated.

## Testing

- Verified on this branch: 42 targeted tests (route guard "blocker wins over every route", status-driven rendering, sign / retry / logout, fail-closed behaviour, login hint); full suite 1111/1111 green.
- `eslint` / `prettier` / `stylelint` clean; `tsc` + `vite build` green; `build-storybook` green.
- All 10 new stories axe-clean against the WCAG 2.2 AA rule set, including the 390×844 long-text story.
- Re-verified at the tip of this stack: full vitest suite green (see the trailing PRs in the stack for the exact counts).
- Clean-slate end-to-end chain run (all services rebuilt from these branches) exercised this gate as scenario **S4**: DPA blocker on login, deep links blocked, signing unlocks durably. Continuity checks **C1–C6** additionally confirmed that an onboarded admin logs in unblocked, that acceptance is persisted verbatim and append-only, that a directly created tenant can read *and* sign the governing operator DPA, and that forwarding is unregressed.

Part of OpenResilienceInitiative/ORISO-Admin#569
Closes OpenResilienceInitiative/ORISO-Admin#572


Shirloin:
Core DPA blocker (#588) and TOC navigation (#590) are both merged to pre-dev. Awaiting QA External Dev verification on Dev before closing.

Issue #661, opened 2026-08-03
Storybook icon catalog with usage cross-reference


kiodreambau:
Affected repos: OpenResilienceInitiative/ORISO-Admin

## Why

Same gap as the Frontend: the Admin panel has ~160 icons with no inventory of what is used. A rescued WIP branch (`save/admin-icon-catalog-wip`) contains the same working Storybook icon-catalog pattern as the Frontend one.

## What to build

Land the Admin Storybook icon-catalog page with its usage cross-reference generator, mirroring the Frontend implementation. Decide committed-vs-gitignored for the generated JSON and document regeneration.

## Acceptance

- [ ] Admin Storybook renders the icon catalog with search and app-used filtering.
- [ ] The generator run is reproducible and documented.
- [ ] No untracked generated output is left behind.
- [ ] `tsc --noEmit` and the Storybook build stay green.

Storypapst:
Verified as delivered: PR #662 (merged into pre-dev on 2026-08-06) ships the Storybook "Design System/Icons" page with search and app-used filtering, plus the deterministic generateIconCatalog.mjs generator and committed catalog JSON; PR #712 (merged on 2026-08-18) fixes catalog drift and adds a CI drift check, fully covering issue #661's acceptance criteria.

Moving to **QA external Dev** for external review. cc @BjoernLudwig

PR #712, opened 2026-08-12
fix(storybook): regenerate stale icon catalog; ignore storybook-static in lint gates

### LINKED ISSUE #711: Stale committed icon catalog produces an unrelated diff on every Storybook build
Refs OpenResilienceInitiative/ORISO-Admin#661

## Why

`src/components/IconCatalog/iconCatalog.generated.json` is stale on `pre-dev`. Commit c000ba0 (mobile FAB nav, #681) added `src/resources/img/svg/nav-glyphs/*` and `src/components/NavGlyph/index.tsx` without re-running the generator. Because `prestorybook` and `prebuild-storybook` both run `npm run generate:icon-catalog`, every contributor who starts or builds Storybook now gets an unrelated 132-line diff in `git status` and has to remember to revert it before committing.

Concretely: `logout.svg` is committed as `usage: "catalogued-only"` with empty `usageFiles`, but the generator finds it used in `src/components/NavGlyph/index.tsx`; the nine `nav-glyphs/` assets are missing entirely.

## What

- Regenerate and commit `iconCatalog.generated.json`.
- Settle committed-vs-gitignored for the generated JSON and write the reasoning down — this is still an open acceptance criterion on #661.
- Separately: `npx eslint .` and `npx prettier . --check` both fail whenever a `storybook-static/` build output exists in the working tree. `.gitignore` ignores it, but `.eslintignore` and `.prettierignore` do not, so both gates walk the 17 MB / 538-file Storybook bundle. Since running `npm run build-storybook` is what produces that directory, the documented validation sequence cannot be run cleanly back-to-back today.

## Goal

`npm run build-storybook` on a fresh `pre-dev` checkout leaves `git status` clean, and the four validation gates (`npm run test`, `npx eslint . --max-warnings=0`, `npx prettier . --check`, `npm run build`) pass with `storybook-static/` present.

Acceptance criteria:

- [ ] `npm run generate:icon-catalog` produces no diff on `pre-dev`.
- [ ] The committed-vs-gitignored decision is stated with its rationale.
- [ ] `npx eslint . --max-warnings=0` and `npx prettier . --check` pass with `storybook-static/` in the tree.

Storypapst:
Closes #711 · Refs #661

## Problem

`iconCatalog.generated.json` on `pre-dev` predates c000ba0 (#681), which added `nav-glyphs/*` and `NavGlyph/index.tsx` without re-running the generator. Since `prestorybook` and `prebuild-storybook` both run `generate:icon-catalog`, everyone who starts or builds Storybook got an unrelated 132-line diff to remember to revert.

## Changes

- Regenerate the catalog: nine `nav-glyphs` entries added, `logout.svg` flips `catalogued-only` → `app-used`. Unedited generator output; consecutive runs are byte-identical.
- **Kept committed, not gitignored.** `node-build` runs ESLint, Prettier, tests and the app build *before* `build-storybook` ever generates the file. Without it committed, `npx eslint .` fails at `IconCatalog.stories.tsx:13` with `import/no-unresolved` (verified by deleting it). Gitignoring would need CI reordering plus a lint exemption — more moving parts than the occasional regeneration it saves.
- **Drift check in `node-build`:** regenerate, then `git diff` the catalog. One definition covers the feature-branch, PR and main workflows, so it guards `pre-dev` directly. Runs before the slower gates — 751 ms in CI — so a stale catalog fails in seconds rather than minutes.
- Add `storybook-static` to `.eslintignore` and `.prettierignore` (the reported issue is `.eslintignore`, not a flat config). `.gitignore` covers it, these did not, so after a local Storybook build ESLint errored on 339 bundled JS files (`that TSConfig does not include this file`) and Prettier flagged 428. CI never hit it, because it lints before building Storybook.
- Carries d69940fe, the stylelint autofix that unblocks any branch merging current `pre-dev`.

## Verification

- Drift check, four cases: current tree → 0; stale catalog → 1 with the `::error::` annotation; dirty `yarn.lock` alone → 0; both dirty → 1 naming only the catalog. Green in CI (751 ms)
- `npm run test` → 1458 passed (237 files)
- `npx eslint . --max-warnings=0` → clean (17s; without the ignore it had not finished after 10 min)
- `npx prettier . --check` → clean
- `npm run build` → ok
- `npm run build-storybook` → ok, `git status` clean afterwards
## Note for a separate issue

The drift check's first revision used a bare `git diff` and went red on its own CI run — on `yarn.lock`, not the catalog. This repo tracks **both** `package-lock.json` and `yarn.lock`, and `npm ci` rewrites the stray `yarn.lock` for the host platform: the committed copy carries `darwin-arm64` entries, the Linux runner resolves `linux-x64`. So the CI working tree is never actually clean, which is invisible today because nothing diffs it. The check is now scoped to the catalog and unaffected, but the duplicate lockfile is worth deciding on separately — `package.json`'s `lint` script still shells out to `yarn run`, so it is not simply removable.

## Summary by CodeRabbit

* **Style**
  * Improved stylesheet formatting and readability.
  * No visual or functional changes were made.

* **Chores**
  * Added automated validation to keep the application’s icon catalog synchronized.
  * Builds now identify outdated icon resources earlier, helping prevent inconsistencies from reaching releases.


kiodreambau:
Conflict resolution + re-verification against current `pre-dev`.

**What conflicted:** only `.prettierignore`. Both sides appended to the same block — this branch added `storybook-static`, `pre-dev` added the `.claude/*` entries. `pre-dev` had meanwhile picked up `storybook-static` by another route, so the resolved file is byte-identical to `pre-dev` and this branch's remaining contribution is the `storybook-static` line in `.eslintignore`.

**Generated artifact:** `iconCatalog.generated.json` was not hand-merged. I re-ran `npm run generate:icon-catalog` after the merge, which picked up the icons `pre-dev` added since this branch was cut (+16 lines on top of the merge result). Re-running `prebuild-storybook` afterwards produced no further diff, so the committed catalog is current — which is the point of this PR.

**Unrelated breakage carried in:** `pre-dev` is currently red on its own `CI - Main` (run 31968762048) — the DPIA merges (#736/#744/#745) landed two `.scss` files that fail `rule-empty-line-before` / `comment-empty-line-before`, so `npm run lint:css` exits 2 for every branch that merges `pre-dev` in. I applied the repo's own `stylelint --fix` to those two files in a separate commit so this branch's gate can go green. That fix belongs on `pre-dev` directly; once it lands there this commit becomes a no-op.

**Verified locally** (Node 26 on macOS; CI uses 22.12.0), running the exact commands from `.github/actions/node-build`:
- `npx eslint . --max-warnings=0` — exit 0
- `npm run lint:css` — exit 0 (`0 errors, 1513 warnings`)
- `npx prettier . --check --ignore-unknown` — "All matched files use Prettier code style!"
- `npm run build` — exit 0
- `npm run build-storybook` — exit 0
- `npm test` — 8 failures locally, all `localStorage`/`sessionStorage` spy tests. I reproduced every one of them on a pristine `origin/pre-dev` checkout with the same local Node, so they are a Node 26 artifact (native `localStorage` cannot be spied on), not a regression from this merge. CI on Node 22 confirms: `build` is green on all three jobs.

Issue #695, opened 2026-08-12
Invite progress view: summary tiles, phase stepper and M3 table on the Links invite tabs


Storypapst:
**Affected repos:** ORISO-Admin

### Why
The invite lists on the Links page (`AccountInvitesTab`, tenant + counsellor tabs) render a plain AntD table with a single status tag. The responsible person at a provider (platform admin tracking invited Träger; tenant admin tracking invited Berater) cannot see where an invitee is stuck in onboarding or what to do about it. This sub-issue redesigns those views in place using the new M3 table family (#621 Package H sub-issue).

### What to build
- Summary strip: four stat tiles (invited, in progress, completed, expired/problem); clicking a tile filters the table.
- Phase-progress column: compact horizontal stepper per invitee. Phases derived **only from data the API already returns** (`AccountInviteStatus`, e-mail delivery state, counsellor provisioning state) via one pure, unit-tested `derivePhases` function; unknown phases render as pending so the component already supports the full phase list (tenant: invited → registered → DPA → 2FA → done; counsellor: invited → account created → done for now). Expired/revoked rows use the magenta error treatment.
- Meta columns (invited at, last activity, status), filter chips, pagination, search where present.
- Row actions preserved from the current tab (resend, copy link, revoke) — no new API calls.
- Empty state with invite CTA; mobile collapse to stacked cards.
- i18n `de`/`en` for all new labels.

### Acceptance
- [ ] Both invite tabs render with the new table family; all previously available actions still work.
- [ ] Stat tiles match the table contents and filter on click.
- [ ] `derivePhases` is unit-tested for every `AccountInviteStatus` value.
- [ ] Expired/revoked rows use the error (magenta) treatment; nothing is hidden, only disabled where not applicable.

### Reviewer test plan
- [ ] Open Links → tenant invites: each row shows a phase stepper; a freshly created draft shows only the first phase filled.
- [ ] Click the "Abgelaufen/Problem" tile → the table shows only expired/revoked/bounced rows; click again to clear.
- [ ] Resend + copy link + revoke from a row → same behaviour as before the redesign.
- [ ] Open the tab at 320px and 412px → rows appear as stacked cards with the mini stepper, all actions reachable.
- [ ] Compare as platform admin vs. tenant admin → each sees only their scope, labels correct in German.

Storypapst:
[...]

Issue #694, opened 2026-08-12
Package H: M3 data-table component family (DataTable, header, pagination, toolbar)


Storypapst:
**Affected repos:** ORISO-Admin

### Why
Epic #621 names "Package H — Table" as the largest single behavioural risk of the AntD removal and asks for its own decision. The three existing table wrappers (`ListingTable`, `EditableTable`, `ResizableTable`) are parallel AntD `Table` wrappers with Storybook stories but no shared canonical component (#276), so table-based views cannot be designed or optimized in Storybook. The invite progress redesign (#309 sub-issue) needs a real M3 table now.

**Decision (in lieu of a separate ADR):** build a custom composable M3 table from the existing design-system atoms — no MUI DataGrid dependency. Rationale: full token control (muted M3 scheme), no license/bundle weight, and our tables need composition (row expansion, phase steppers, chip cells) more than spreadsheet features. Revisit only if virtualisation for very large datasets becomes a requirement.

### What to build
Component family under `src/components/DataTable/`, each part with its own story and vitest coverage:
- `DataTable` — semantic table, M3 tokens, sticky header, row hover, row-expansion slot, loading skeleton, empty-state slot.
- `DataTableHeader` — sortable column headers with accessible `aria-sort`.
- `DataTablePagination` — rows-per-page + page navigation, i18n `de`/`en`.
- `DataTableToolbar` — slot row for filter chips, search, actions.
- Mobile behaviour: rows collapse to stacked cards below the tablet breakpoint.

### Acceptance
- [ ] All components exist with stories; `npm run build-storybook` passes (stories run as CI component tests).
- [ ] Keyboard + screen-reader accessible (sort buttons, pagination controls, `aria-sort`).
- [ ] No new runtime dependency added.
- [ ] At least one production view consumes the family (see invite progress sub-issue of #309).

### Reviewer test plan
- [ ] Open Storybook → `DataTable` stories render in light and dark theme with the muted M3 look, matching existing cards.
- [ ] Sort a column via keyboard only → order changes and the active sort is announced/visible.
- [ ] Switch a story to 320px viewport → rows collapse into readable stacked cards.
- [ ] Pagination: change rows-per-page and navigate pages → counts and controls stay consistent.

Storypapst:
[...]

PR #766, opened 2026-08-15
fix(links): guard stale invite loads and prune selection on filter change

Storypapst:
Refs OpenResilienceInitiative/ORISO-Admin#695, OpenResilienceInitiative/ORISO-Admin#694

## Why
Follow-up to #700 (merged). CodeRabbit's review round landed after the merge, so these fixes need their own PR. Two of the nine findings were real defects, one of them user-visible.

## What
**Stale invite loads could overwrite fresher data (major).** `loadInvites` runs from the mount effect and from `onCreate`, `onResend`, `onRevoke` and the CSV import, and each run walks several pages — so a slower older run really can land last and replace newer state. A revision counter now gates the commit: only the newest run may call `setInvites`, raise the error toast, or clear the loading state. Two tests drive genuinely overlapping loads through the UI with deferred promises; both go red with the guard removed.

**A filtered-out row stayed selected and could be acted on unseen (major).** `selectedInvites` filtered against the full invite list rather than the rendered rows, so a checkbox ticked before a filter change survived the filter and would be included in a bulk resend or revoke without the user seeing it. `AccountInvitesTab` already documents pruning-on-reload as the intended contract, so the tile and chip filters now share one `matchesFilter` predicate and prune the selection to the rows the new filter still shows.

**Seven smaller fixes.** `aria-selected` removed from rows (invalid inside `role="table"`; the row checkbox is what announces and toggles selection); a sortable column without `onSortChange` degrades to a plain header instead of rendering a dead button (five stories were doing exactly that); pagination clamps an out-of-range page so a shrunk list can no longer render "21–4 von 4"; the board writes the clamped page back to state; `twoFactorWaivedAt` counts as activity; keyboard activation (Enter/Space) is now covered; brittle DOM-structure assertions replaced by behaviour assertions; deprecated `clip` declaration removed.

## Notes
- CodeRabbit claimed the deprecated `clip` fails the Stylelint gate. It does not: this repo pins Stylelint 14.16.1, which has no `property-no-deprecated` rule. Removed anyway as correct CSS.
- `npm run lint:css` is not usable as a gate — it runs `stylelint --fix`, and on a clean `pre-dev` checkout it rewrites ~95 files and still reports 261 pre-existing errors. Verified instead that the touched SCSS introduces no new errors against the `pre-dev` baseline. Worth its own ticket.

## Gates
`npm run test` 267 files / 1770 tests, eslint (touched, `--max-warnings=0`), prettier, `typecheck:storybook`, `build`, `build-storybook` — all green.

## How should I test this?
- [ ] Links → invite tab: create an invite, then immediately trigger a resend → the list always ends on the newest data, never flickers back to an older state.
- [ ] Tick a row, then switch the status filter so that row disappears → run a bulk action: the hidden row is NOT included.
- [ ] Filter down until only one page remains while you are on page 2 → the range reads sensibly and navigation stays inside bounds.
- [ ] Keyboard only: focus an expandable row, press Enter and Space → the row expands; tab order stays sane.
- [ ] Screen reader on the table → selection is announced via the row checkbox, not a phantom row state.

## Summary by CodeRabbit

* **Accessibility**
  * Improved data table keyboard navigation and selection announcements.
  * Excluded non-interactive rows from the tab order.
  * Updated visually hidden styles for standards compliance.

* **Bug Fixes**
  * Prevented invalid pagination after result counts shrink.
  * Ensured overlapping invite loads display only the latest results.
  * Preserved selection and pagination when filtering or result sets change.
  * Improved invite activity tracking for two-factor authentication waivers.
  * Sort indicators now appear only when sorting is available.


shazia-k (approved):
Real fixes, tight patches, and I like that the tests actually harness the bug (the deferred-promise pair for the stale-load race is much better than a `waitFor` guess). Specifically: the revision guard is done in all three branches (success, error, finally) so a superseded run leaves the `loading` flag to whoever overtook it — good. `applyFilter` extracting the shared `matchesFilter` predicate means the pruning and the rendering can't drift, and the page clamp is written back via `useEffect` so a shrink-then-grow doesn't snap back to a stale page. Dropping `aria-selected` on `DataTableRow` is the correct call (rows on plain `role="table"` don't support it), and degrading `sortable` columns to plain when no handler is wired is the right a11y guardrail. Approving — please rebase, it's currently CONFLICTING.

Issue #609, opened 2026-07-30
Agency/department DPP editor has no role/delegation permission gating (frontend or backend)


Storypapst:
## Context

Found while verifying #41 (role-aware settings). The tenant-level data privacy policy editor correctly gates edit rights behind a delegable Super-Admin toggle (`legalContentChangesBySingleTenantAdminsAllowed`), enforced both client-side (`readOnly`) and server-side (`TenantFacadeAuthorisationService.isAllowedToEditLegalContent`, 403 on violation). The **agency/department-level** DPP editor has no equivalent gating at all.

## Problem

- Frontend: `AgencyLegalTextContainer` → `DepartmentDataProtectionCard` (`src/pages/Agency/Edit/index.tsx:408-418`) renders the DPP editor always-editable — no role/permission check, no `readOnly` state.
- Backend: `AgencyAdminController.publishDepartmentDataProtection` (line ~350) has no `@PreAuthorize` and no delegable-right check. `DepartmentDataProtectionService` only enforces agency/tenant *ownership* (IDOR guard) — any admin who owns the agency can always edit and publish, regardless of role or any platform/tenant-level toggle.

## Ask

Mirror the tenant-level pattern at the agency/department level:
- A platform/tenant-level toggle (or the natural agency-level equivalent) that decides whether a Counselor Admin can edit their own department's DPP.
- Client-side: render read-only when not granted.
- Server-side: reject the publish/update call when not granted, not just rely on ownership.

## Repos affected

`ORISO-Admin` (frontend gating), `ORISO-AgencyService` (backend enforcement).

Found via code audit during #41 (ORISO-Admin), not live-tested — worth a quick manual confirmation before fixing.

Storypapst:
Frontend half is in https://github.com/OpenResilienceInitiative/ORISO-Admin/pull/779.

The Fachbereich data-protection and imprint editors now render read-only for an admin without the legal-text right: editor inert, no publish, no draft-save, consent sentence and per-field translate disabled. `UserRole.AgencyAdmin` gained a `LegalText` entry mirroring the tenant-level rule — reading always, changing behind `singleCanEditLegalText`. In a deployment without single-domain multitenancy that flag is `true`, so nothing changes there.

## Two things this issue asks for that need a decision before they can be built

[...]

PR #779, opened 2026-08-16
Gate the Fachbereich legal editors behind the legal-text right (#609)

Storypapst:
## Problem

The Fachbereich data-protection and imprint editors had **no permission check at all**. `AgencyLegalTextContainer` and the department containers rendered the editor always-editable, and the agency-admin role carried no `LegalText` entry — so nothing ever asked. Any admin who owned the agency could publish a data-protection policy regardless of role or delegation.

## What changed

- `UserRole.AgencyAdmin` gains `LegalText: { read: true, update: singleCanEditLegalText }` — the tenant-level rule one rung down the ladder. Without single-domain multitenancy that flag is `true`, so default deployments behave exactly as before.
- `DepartmentDataProtectionCard` takes `readOnly`: editor inert, no publish, no draft-save, consent sentence and per-field translate disabled. Disabled rather than hidden, so a viewer still sees which document the Fachbereich has.
- `AgencyLegalTextContainer`, `DepartmentDataProtectionContainer` and `DepartmentImprintContainer` pass it from `can(Update, Resource.LegalText)`.

## Deliberately not in this PR

**Server-side enforcement.** ORISO-AgencyService has no legal delegation toggle today — no `legalContentChangesBySingleTenantAdminsAllowed`, no equivalent; `DepartmentDataProtectionService` enforces agency/tenant *ownership* only. Where that switch belongs (platform setting vs. the ADR-013 agency admin controls) is a product decision, raised on #609.

**Restricted agency admins.** They inherit the new right through `UserRole.AgencyAdmin` and are not capped by `enforceRestrictedAgencyAdminCeiling`. That matches today's behaviour (they could already edit unconditionally); whether they should is a separate question, also raised on #609.

## Verification

- `npm run test` — 276 files, 1830 tests, green
- `npx eslint . --max-warnings=0`, `npx prettier . --check`, `npm run build` — green
- Both gates mutation-checked: removing the permission row fails 4 assertions, ignoring `readOnly` fails 2.

Refs #609

### DIFF
--- a/src/components/Tenants/LegalSettings/components/AgencyLegalTextContainer/AgencyLegalTextContainer.test.tsx
+++ b/src/components/Tenants/LegalSettings/components/AgencyLegalTextContainer/AgencyLegalTextContainer.test.tsx
@@ -6,6 +6,7 @@ const h = vi.hoisted(() => ({
     useDepartmentDpp: vi.fn(),
     refetch: vi.fn(),
     card: vi.fn(),
+    canEditLegalText: vi.fn(() => true),
 }));
 
 vi.mock('react-i18next', () => ({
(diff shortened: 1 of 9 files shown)


Storypapst:
@coderabbitai review

Issue #607, opened 2026-07-30
Design review: invite composer, template dialog, DPA screen and shared M3 primitives


Storypapst:
## Why

An owner review of the Links section and the DPA blocker found the same classes of defect across several screens. Several of them live in shared primitives, so they show up wherever those are used: off-scheme colours from three different palettes, a split button that always claimed the elevated state, chips that read as traffic lights, missing icons and English fallbacks in a German UI.

## What

Work through the annotated screenshots as one pass, fixing the shared component where the defect lives rather than patching each page.

## Goal

- The Links section and the DPA blocker match the reviewed Figma frames (1611-27868 for the AVV screen, 1165:16407 for the split buttons).
- Every colour resolves from the ORISO M3 tokens — no navy, black or Tailwind greys.
- Elevation is a state, not a skin: a split button lifts only while its menu is open.
- Nothing in the German UI falls back to English.
- A template can be picked from the overview list, the CSV import explains its format, and dialogs say what they are for.
- Validation gates green: `npm run test`, `npx eslint . --max-warnings=0`, `npx prettier . --check`, `npm run build`.

Storypapst:
Partial delivery: #781, merged into `pre-dev`.

Two findings that had been merged unaddressed with #765 are fixed:
- the M3 editor rollout page claimed the editor "expects a language map" — it takes a single HTML `value`; the host owns the map. Stated wrongly, the note would have deterred exactly the single-value candidates the page lists;
- its verdict chips used `--m3-on-secondary-container` as a **background** with hardcoded `#000` text. Each chip now pairs a container token with its matching `on-` token, and the presentation moved out of inline styles into a SCSS module on the shared `$mui-unit` scale.

The rest of this issue — Links section, DPA blocker screen against Figma 1611-27868, split-button elevation against 1165:16407, chips reading as traffic lights, English fallbacks in the German UI, template picking from the overview list, CSV import copy — is untouched and still open.

shazia-k:
## Status audit 2026-08-28 — need @Storypapst clarification before this can close

Two merged PRs cite this issue, but they tell inconsistent stories:

**#608 (MERGED 2026-07-30) — Storypapst's own PR description claims full delivery:**
- Colours onto scheme (navy `#273270`, black `#1f1f1f`, Tailwind greys → ORISO M3 tokens; error role completed; tonal chips)
[...]

Issue #583, opened 2026-07-28
Fachbereich switcher does not show which departments left the inherited text


kiodreambau:
## Problem

The Fachbereich switcher in the agency legal-text editors (#555, PR #562) does not mark which
departments have already left the inherited text. An admin editing "Alle Fachbereiche" therefore
cannot see who will *not* receive the change — the one thing that matters when a Beratungsstelle
publishes a correction.

CodeRabbit raised this twice on #562 and I declined it with a rationale that is wrong: I wrote that
the backend offers no bulk signal. It does. `GET /service/agencies/{id}` has carried
`departments[].hasPublishedDpp` / `hasPublishedImprint` since ORISO-AgencyService `8bb7578`, which
is live on `pre-dev`.

## Why it is still not implemented

That signal is on the **public** agency read. The admin page uses
`GET /service/agencyadmin/agencies/{id}`, whose `AgencyAdminResponseDTO` has no `departments` list.
Reaching across to the public endpoint from the admin UI just to render a marker is the wrong shape.

## What to build

1. `ORISO-AgencyService`: add the same `departments[]` (topicId + `hasPublishedDpp` +
   `hasPublishedImprint`) to `AgencyAdminResponseDTO`, populated by the existing
   `convertToAgencyDepartmentDTO` logic in `AgencyService`.
2. `ORISO-Admin`: render the marker in `DepartmentSelect`, and remove the "tracked in #567" note
   from its doc comment.

## Acceptance

- The switcher shows, per Fachbereich, whether it carries its own published text.
- A department the admin has never opened is marked correctly — the marker must not depend on what
  happened to be fetched.

Refs OpenResilienceInitiative/ORISO-Admin#562
Refs OpenResilienceInitiative/ORISO-AgencyService#212
Part of epic OpenResilienceInitiative/ORISO-Frontend#86

Storypapst:
Delivered in #780 (Admin) and OpenResilienceInitiative/ORISO-AgencyService#259 (backend), both merged into `pre-dev`.

**Backend:** `AgencyAdminResponseDTO` now carries `departments[]` (`topicId`, `hasPublishedDpp`, `hasPublishedImprint`), built from the already-loaded `agencyTopics` association — no extra query. The ADR-014 resolution order moved out of `AgencyService`'s private helpers into a shared `DepartmentLegalPublicationState`, so the admin read and the public read cannot drift; a switcher that disagrees with what help-seekers see would be worse than no marker.

[...]

PR #780, opened 2026-08-16
Mark which Fachbereiche have left the inherited legal text (#583)

Storypapst:
## Problem

The Fachbereich switcher in the agency legal editors does not mark which departments have already left the inherited text. An admin editing "Alle Fachbereiche" therefore cannot see who will **not** receive the change — the one thing that matters when a Beratungsstelle publishes a correction.

## What changed

- `DepartmentSelect` marks each Fachbereich that carries its own published text, and tells the "Alle Fachbereiche" entry how many that is — the number is the actual decision input when writing a correction.
- `AgencyLegalTextContainer` maps `departments[]` from the agency read onto the switcher **per kind**: the privacy editor reads `hasPublishedDpp`, the imprint editor `hasPublishedImprint`. A Fachbereich with its own imprint may still inherit the policy, so one flag for both would state the opposite of the truth.
- Where the backend does not report the state, entries carry **no** marker rather than a wrong one. A missing marker must not read as "still inherits".
- The chip is a neutral outline, not a status colour: having your own text is a fact about inheritance, not a problem. Traffic-light chips were called out in the design review (#607).

## Depends on

**ORISO-AgencyService#259**, which adds `departments[]` (topicId + hasPublishedDpp + hasPublishedImprint) to `AgencyAdminResponseDTO`. Until that ships this degrades to exactly today's behaviour — no markers, everything else unchanged. Because the state comes with the agency read, a department the admin never opened is marked correctly, which the acceptance criteria require.

## Verification

- `npm run test` — 274 files, 1820 tests, green
- `npx eslint . --max-warnings=0`, `npx prettier . --check`, `npm run build` — green
- Mutation-checked: making the mapping ignore `field` fails 1 assertion; never rendering the marker fails 3

Refs #583, OpenResilienceInitiative/ORISO-AgencyService#259

## Summary by CodeRabbit

* **New Features**
  * Department selectors now indicate which departments have their own privacy or imprint text.
  * Agency-wide selections show how many departments will not inherit updated legal text.
  * Added English and German labels for legal-text status indicators.

* **Bug Fixes**
  * Legal-text ownership indicators now reflect the selected topic accurately.
  * Missing status information no longer displays misleading indicators.

* **Tests**
  * Expanded coverage for inherited text, department-specific text, counts, and department switching.


Storypapst:
@coderabbitai review

Storypapst:
Correction pushed: the first commit wrote the new marker tests **into** `DepartmentSelect.test.tsx`, replacing the five behaviour tests that were already there. Both sets are now in the file (10 tests). Nothing else changed.

PR #784, opened 2026-08-16
fix(lint): get pre-dev green — two stylelint errors, two different right answers

kiodreambau:
**`pre-dev` is red on its own `CI - Main`.** `npm run lint:css` exits non-zero, so every branch that merges `pre-dev` inherits the failure — the five Admin PRs repaired today each had to carry an identical carry-fix just to go green. This fixes it at source; those carry-fixes become no-ops.

| commit | `CI - Main` |
|---|---|
| `0fa59803` (17:46) | ✅ green |
| `4502e00c` (19:49) | ❌ red |
| `cb6e8a59` (HEAD) | ❌ red |

Introduced by the DPIA merges (#736 / #744 / #745).

## Three errors, two rules, two different right answers

This is deliberately **not** a blanket `stylelint --fix`.

### 1. `rule-empty-line-before` — config, not content

`src/components/Dpia/styles.module.scss:319` and `src/pages/GlobalSettings/styles.module.scss:16` are both **multi-line doc comments sitting directly above the rule they document**:

```scss
/** Chapters not yet ported into this view (see AVAILABLE_CHAPTER_IDS) — shown
    for orientation but not a link, so there is nothing for the anchor to fail
    to reach. */
.chipDisabled {
```

The config already carries `"after-single-line-comment"` in its `except` list, so the intent is established: *a comment immediately above a rule documents it and should not be forced apart from it.* That exception simply does not cover multi-line comments.

**`--fix` here would insert a blank line BETWEEN the doc comment and the rule it describes** — divorcing the explanation from its subject, the exact opposite of what the existing exception protects.

So the fix is `"after-single-line-comment"` → `"after-comment"`, which subsumes it.

### 2. `comment-empty-line-before` — content, not config

`src/components/Dpia/styles.module.scss:18` is a **different rule**, and here a blank line is genuinely correct: it is a section comment introducing a group of custom properties mid-block, not a doc comment glued to a following rule. Added the blank line.

## Verification

Confirmed `"after-comment"` is a valid stylelint 17 option and that it introduces nothing new — the **warning count is byte-identical** before and after:

| | problems |
|---|---|
| pristine `pre-dev` | `1516 problems (**3 errors**, 1513 warnings)` |
| with this change | `1513 problems (**0 errors**, 1513 warnings)` |

Only the errors change. The 1513 warnings (property order, unsupported browser features) are pre-existing and deliberately out of scope — this restores exactly the error-free state CI was last green on, and nothing more.

Also verified `npm run lint:js` and `npm run lint:formatting` pass.

## Reviewer test plan

- [ ] `CI - Main` / PR build is green
- [ ] `npx stylelint "src/**/*.(s(c|a)ss|css|less)"` reports **0 errors**
- [ ] Agree that widening the exception beats `--fix` splitting doc comments from their rules
- [ ] Confirm the 1513 pre-existing warnings staying untouched is the right scope (they are worth their own ticket)

Note: npm rewrites `yarn.lock` whenever it is present — that rewrite was reverted, so this PR touches only the two intended files.

### DIFF
--- a/src/components/Dpia/styles.module.scss
+++ b/src/components/Dpia/styles.module.scss
@@ -15,6 +15,7 @@
     --dpia-success-ink: #075e26;
     --dpia-success-wash: #e8f3eb;
     --dpia-success-line: #b9dcc2;
+
     /* "Planned/not shipped" amber family, used by the appointment-module card
        and its badges. Named aliases so the shade is declared once instead of
        repeated as raw hex at every call site. */
--- a/stylelint.config.js
+++ b/stylelint.config.js
@@ -78,6 +78,13 @@ module.exports = {
         "except": [
           "after-single-line-comment",
           "first-nested"
+        ],
+        // A comment directly above a rule documents it; forcing a blank line
+        // between them separates the explanation from its subject. "except"
+        // only covers single-line comments, so multi-line doc comments need
+        // this ignore.
+        "ignore": [
+          "after-comment"
         ]
       }
     ],


kiodreambau:
Correction to my own first push — worth reading, because the first attempt was wrong in an instructive way.

**What I pushed first:** moved `"after-single-line-comment"` → `"after-comment"` inside the rule's `except` list. Locally that reported `0 errors` and looked correct.

**It was not correct.** `after-comment` is not a valid value for `except` on `rule-empty-line-before`. Stylelint's response to an invalid option is not to fail loudly — it emits an `invalidOptionWarning`, **silently skips the entire rule**, and still exits 2:

```
Invalid value "after-comment" for option "except" of rule "rule-empty-line-before"
```

So the two errors "disappeared" because **the rule had been switched off**, not because the code satisfied it. That is strictly worse than the failure it replaced: a disabled lint rule that reads as a clean run. CI caught it, and chasing that rather than retrying is what surfaced the real cause.

**What is pushed now:** the option belongs in `ignore`, not `except`. `except` keeps its original valid values.

```js
"rule-empty-line-before": [
  "always-multi-line",
  {
    "except": ["after-single-line-comment", "first-nested"],
    "ignore": ["after-comment"]
  }
]
```

**Verified behaviourally, not by error count** — the point being to prove the rule is still *enforced* rather than merely quiet:

| probe | result |
| --- | --- |
| rule directly after a rule, no blank line | **still reported** ✅ (rule is live) |
| rule after a multi-line doc comment | not reported ✅ (the intended exception) |

Plus: `npm run lint:css` **exit 0** with **no `invalidOptionWarnings`**, `npx eslint . --max-warnings=0` exit 0, `npx prettier . --check` exit 0.

**A note that generalises beyond this PR:** `0 errors` from stylelint is not sufficient evidence that a rule passed — an invalid option produces exactly that output while disabling the rule. The reliable signal is `--formatter json` and checking `invalidOptionWarnings` is empty. Given this repo's lint gate is a required check, that distinction matters.

kiodreambau:
Superseded by #786, which fixes the same three errors without changing `stylelint.config.js`.

The diagnosis here is right, and so is the reasoning about doc comments: forcing a blank line between a comment and the rule it documents separates the explanation from its subject, and that is worth avoiding. The difference is only in the mechanism.

`ignore` is evaluated before `except` in `rule-empty-line-before`, so `ignore: ["after-comment"]` does not just cover multi-line doc comments — it also stops enforcing the existing `except: ["after-single-line-comment"]`. Probe file:

```scss
.a {
    color: red;
}

// single-line doc comment, then a blank line before the rule

.b {
    color: blue;
}
```

| config | result |
|---|---|
| `pre-dev` today (and #786) | `✖ Expected no empty line before rule` — exit 2 |
| this branch | clean — exit 0 |

Both PRs are green in CI, so this difference would not surface at review time.

#786 reaches the same readability outcome through the `//` comment form, which is what `except: ["after-single-line-comment"]` exists to permit, and what `src/components/Dpia/styles.module.scss` already does for the comment above `.section`:

```scss
// The app bar is no longer sticky at this width, so only the (still sticky) chapter nav's
// own height needs to be cleared when jumping to a chapter via anchor.
.section {
```

The `comment-empty-line-before` half of this PR — the blank line in the Dpia custom-property block — is identical in #786; that one genuinely is a group separator rather than a doc comment, so the plain fix is correct there.

Closing in favour of #786. Nothing in the analysis here was wrong, and its write-up of the two-rules-two-answers split is what prompted the closer look at the `ignore`/`except` ordering.

kiodreambau:
Closing in favour of #786, which fixes the same three stylelint errors with a better-scoped change.

Both PRs are green and both restore `pre-dev`. The difference is blast radius:

| | approach | scope |
| --- | --- | --- |
| this PR | add `"ignore": ["after-comment"]` to `rule-empty-line-before` in `stylelint.config.js` | **loosens the rule repo-wide** for every multi-line comment |
| #786 | rewrite the two `/** */` doc comments as `//` so they match the config's existing `after-single-line-comment` exception | **2 files, no config change** |

#786 is the better fix. It does not weaken a lint rule across the whole repository to resolve two call sites, and it works with the configuration's existing intent rather than widening it. Both also add the same blank line for the separate `comment-empty-line-before` error, so #786 is complete on its own.

## One finding from this PR worth keeping

The first version of this branch put `after-comment` in the rule's **`except`** list rather than `ignore`. That is not a valid value, and stylelint's response to an invalid option is not to fail loudly:

```
Invalid value "after-comment" for option "except" of rule "rule-empty-line-before"
```

It emits an `invalidOptionWarning`, **silently skips the entire rule**, and still exits 2. Locally that reads as `0 errors` while the rule has actually been switched off — strictly worse than the failure it replaced, on a required check.

**`0 errors` from stylelint is therefore not sufficient evidence that a rule passed.** The reliable signal is `stylelint --formatter json` and confirming `invalidOptionWarnings` is empty. Worth knowing for anyone touching this config in future.

The underlying cause of the red `pre-dev` is unchanged and documented in OpenResilienceInitiative/ORISO-Docs#89: merging to `pre-dev` runs a weaker gate than the PR did, so a bad merge lands unblocked and then taxes every open PR in the repo.

PR #704, opened 2026-08-12
chore(deps): Dependabot triage - remove noise, fix real CVEs, add config

Storypapst:
Dependabot triage: cut the noise, fix what actually ships.

Noise removal:
- Delete yarn.lock (388 KB). The repo builds with npm: every workflow uses
  `npm`, no workflow or Dockerfile mentions yarn, and the image copies a
  prebuilt `build/` directory. The file was a second, independently resolved
  dependency tree, so GitHub reported every finding twice - 47 of this repo's
  ~95 open alerts were duplicates of the package-lock.json ones.
- Fix the `lint` script, which shelled out to `yarn run` and therefore failed
  for anyone without yarn installed, and drop the now-dead .prettierignore entry.

Security fixes (package-lock.json, via `npm audit fix`, no breaking changes):
- 18 -> 9 npm advisories. dompurify 3.4.13 (the XSS sanitizer),
  linkify-it 5.0.2, brace-expansion 1.1.18, minimatch 3.1.5, and transitives.

Deliberately NOT included - these need their own tickets, not a bot PR:
- axios 0.25.0 -> 1.x, immutable 3.x -> 5.x, react-router 7 -> 8: all major
  upgrades with real breaking-change surface.
- lodash.set: prototype pollution, no fixed version exists (package is
  deprecated); needs replacing, not bumping.

Configuration:
- Add .github/dependabot.yml: monthly cadence, security PRs grouped and split
  by production vs development dependency type so production findings stay
  visible, majors excluded. See the comment in that file on why security PRs
  still land on `main`.

Verified: npm ci, npm run build, and npm run test all green
(236 test files, 1456 tests passing) on Node 22.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Added automated dependency update checks for npm packages, GitHub Actions, and Docker base images.
  * Configured monthly update scheduling and grouped pull requests for easier maintenance.
  * Added separate handling for security updates and major-version upgrades.
  * These changes help keep application dependencies and development tooling current while reducing maintenance overhead.

PR #702, opened 2026-08-12
fix(invites): mail glyph on "Direkt Versenden", filled once it can fire

Storypapst:
## Problem

The invite composer's send button carried a paper plane in every state, so the one thing the two send modes differ in — whether an e-mail actually leaves the building — was invisible. "Empfänger nur anlegen" showed a send glyph while sending nothing. Part of #574.

## What changed

- **"Direkt Versenden" carries the mail glyph**: outlined while the action is still gated, filled once it can fire. Its entry in the send-options menu follows the same rule.
- **"Empfänger nur anlegen" takes the file glyph** its own menu entry already carries, instead of the paper plane.
- **`mail_24px` / `mail_filled_24px` lose their Figma `<g id>` wrapper.** Both mail icons now render twice on the page (template picker + send button); the fixed ids would otherwise be duplicated in the DOM — the same shape as the mask-id collision that made glyphs vanish before.
- The glyph is driven by one `renderSendGlyph()` helper, covered by three new tests.

Bulk-selection mode is untouched — it keeps the select-all glyph.

## Verification

- `npx vitest run src/pages/Links src/components/GlobalSearch` — 67 passed (3 new, red before the fix).
- `npx eslint` + `npx prettier --check` on the changed files — clean.
- `npm run build` — passes.
- Storybook `Organisms/Pages/Links/InviteComposer`, states checked by hand: gated (outlined mail), valid (filled mail on the brand-red CTA), create-only (file glyph), send menu open (filled mail marks the selected mode).

## Reviewer test plan

- [ ] Links → Träger tab: the send button shows an **envelope**, not a paper plane.
- [ ] With the form still incomplete the envelope is outlined; filling a valid e-mail flips it to the filled envelope on the red button.
- [ ] Open "Sendeoptionen": "Direkt Versenden" carries the filled envelope, "Empfänger nur anlegen" the file glyph.
- [ ] Switch to "Empfänger nur anlegen": the button glyph becomes the file glyph — no paper plane on a mode that mails nothing.
- [ ] Select rows to enter bulk mode: the select-all glyph is unchanged.
- [ ] Repeat on a phone viewport — the composer row scrolls, so scroll to the end to reach the button.

## Summary by CodeRabbit

* **Improvements**
  * Updated invite actions with clearer icons for sending invitations and saving links.
  * Adjusted icons dynamically based on the selected invite mode and readiness state.
  * Improved the direct-send menu icon for better visual consistency.

* **Tests**
  * Added coverage for invite action icons across supported sending modes.

PR #699, opened 2026-08-12
test(storybook): run every story as a component test in CI

Storypapst:
## Problem

Storybook 10 is configured in this repo, but nothing ever executed the `play` functions. Story-level defects only surfaced when a human happened to open the story.

## What changed

- `@storybook/addon-vitest` + `@storybook/addon-docs`. `vitest.config.ts` is split into a `unit` project (the existing jsdom suite — still what `npm test` runs) and a `storybook` project that renders every story in a real Chromium, runs its `play` function, and fails on any axe violation (`a11y.test: 'error'`, WCAG 2.2 AA).
- New `storybook-tests` job in `ci-pull-request.yml` (Node 22, Chromium only), plus `test:storybook` / `test:storybook:coverage` scripts.
- The test browser is pinned to `de-DE` / `Europe/Berlin`. Headless Chromium reports `en-US` and `src/i18n.ts` takes its language from the browser, so play functions asserting German labels passed for a reviewer in Germany and failed anywhere else.
- Late-discovered dependencies are listed in `optimizeDeps`. Vitest imports stories dynamically, so Vite never sees them during its initial crawl, re-optimizes mid-run and kills every request in flight — which showed up as ~9 unrelated stories failing at random on a cold cache. `DEBUG=vite:deps` identifies them.
- `src/Introduction.mdx`; the coverage output is excluded from ESLint and Prettier.

## Fixed along the way

An id starting with a digit used as a CSS selector (`#4-…` throws `SyntaxError`); one `HttpResponse` instance shared across requests, which made MSW throw "Response body is already used"; two stories asserting English copy.

## Known exclusion

`EditorImageUpload > UploadError` carries `tags: ['!test']` with the reason inline: MSW's service-worker path never settles the non-2xx response, so the component never sees the rejection and no toast renders. The 201 siblings in the same file pass, so this is an MSW limitation, not an editor defect.

## Verification

| Command | Result |
| --- | --- |
| `npm run test:storybook` (cold cache) | 132 files / 478 tests pass, exit 0 |
| `npm test` | 236 files / 1456 tests pass |
| `npx eslint . --max-warnings=0` | pass |
| `npx prettier . --check` | pass |
| `npm run typecheck:storybook` | pass |
| `npm run test:storybook:coverage` | pass, 42.5% lines |

## Reviewer test plan

- [ ] `npm ci && npx playwright install chromium && npm run generate:icon-catalog && npm run test:storybook` — green
[...]

PR #786, opened 2026-08-17
fix(lint): restore the stylelint gate on pre-dev

kiodreambau:
## Problem

`pre-dev` is red on its own `CI - Main` workflow. The `Lint (Stylelint)` step of `.github/actions/node-build` fails because `npm run lint:css` exits 2:

```
✖ 1516 problems (3 errors, 1513 warnings)
```

Example run: https://github.com/OpenResilienceInitiative/ORISO-Admin/actions/runs/31968762048

The 3 errors were introduced by the DPIA merges (#736 / #744 / #745). Because every feature branch that merges `pre-dev` in inherits them, the identical throwaway `stylelint --fix` commit had to be carried into five open PRs (#712, #702, #699, #704, #673) just to get their gates green. This replaces those duplicates with one fix.

## The three errors, and why they get two different treatments

| File | Rule |
|---|---|
| `src/components/Dpia/styles.module.scss:18` | `comment-empty-line-before` |
| `src/components/Dpia/styles.module.scss:319` | `rule-empty-line-before` |
| `src/pages/GlobalSettings/styles.module.scss:16` | `rule-empty-line-before` |

**`comment-empty-line-before` — blank line added.** The comment introducing the amber custom-property group sits flush against the group above it. Here the blank line is the right read: it separates two groups of declarations rather than binding a doc comment to one of them. This is the plain autofix and it is correct.

**`rule-empty-line-before` — switched to the `//` comment form.** Both cases are a multi-line block comment documenting the rule directly beneath it. The plain autofix inserts a blank line *between* the doc comment and the rule it documents, splitting the two things that belong together.

Instead they now use the `//` form, which is exactly what `except: ["after-single-line-comment"]` in `stylelint.config.js` exists to permit — and what the same `Dpia/styles.module.scss` already does for the comment above `.section`:

```scss
// The app bar is no longer sticky at this width, so only the (still sticky) chapter nav's
// own height needs to be cleared when jumping to a chapter via anchor.
.section {
```

`//` comments are also stripped by Sass rather than emitted into the built CSS, which suits an internal doc comment.

## The gate is not weakened

**No change to `stylelint.config.js`.** `rule-empty-line-before` still enforces in both directions — putting a blank line after these comments is now an error, verified:

```
17:1  ✖  Expected no empty line before rule    rule-empty-line-before
```

The `//` form is the shape the rule *requires*, not an exemption from it.

### A note on the shorter route

Adding `"after-comment"` to the rule's `except` list looks like the tidy fix matching the config's intent, but stylelint does not accept it there — only `ignore` takes `after-comment` (`except` accepts `after-rule`, `after-single-line-comment`, `first-nested`, `inside-block-and-after-rule`, `inside-block`). Tried against the pinned stylelint 17:

```
Invalid Option: Invalid value "after-comment" for option "except" of rule "rule-empty-line-before"
⚠ 1513 problems (0 errors, 1513 warnings)
```

Note the `0 errors`: an invalid option makes stylelint skip the rule entirely. That route would have produced a green gate with `rule-empty-line-before` silently disabled repo-wide — a worse outcome than the red gate it fixed.

Using `ignore: ["after-comment"]` is valid but also weakens the gate: `ignore` is evaluated before `except`, so it would stop enforcing the existing no-blank-line-after-single-line-comment rule as well.

## Verification

Run in a clean worktree off `pre-dev` with `npm ci --ignore-scripts`, matching the CI step:

- `npm run lint:css` → **exit 0**, `⚠ 1513 problems (0 errors, 1513 warnings)`
- `npx prettier . --check --ignore-unknown` → **exit 0**, `All matched files use Prettier code style!`

Warnings go 1516 → 1513: exactly the three errors removed, no new warnings introduced. The remaining 1513 are pre-existing and non-blocking (`severity: "warning"`).

## Follow-up for the five PRs

Once this merges, the `style(lint): restore the stylelint gate broken on pre-dev` commit in #712, #702, #699, #704 and #673 becomes redundant. Those branches should drop it and merge `pre-dev` in instead — if both survive, the two versions of the `rule-empty-line-before` fix will conflict in the same lines.

### DIFF
--- a/src/components/Dpia/styles.module.scss
+++ b/src/components/Dpia/styles.module.scss
@@ -15,6 +15,7 @@
     --dpia-success-ink: #075e26;
     --dpia-success-wash: #e8f3eb;
     --dpia-success-line: #b9dcc2;
+
     /* "Planned/not shipped" amber family, used by the appointment-module card
        and its badges. Named aliases so the shade is declared once instead of
        repeated as raw hex at every call site. */
@@ -313,9 +314,9 @@
     }
 }
 
-/** Chapters not yet ported into this view (see AVAILABLE_CHAPTER_IDS) — shown
-    for orientation but not a link, so there is nothing for the anchor to fail
-    to reach. */
+// Chapters not yet ported into this view (see AVAILABLE_CHAPTER_IDS) — shown
+// for orientation but not a link, so there is nothing for the anchor to fail
+// to reach.
 .chipDisabled {
     border-color: var(--dpia-hair-soft);
     color: var(--m3-on-surface-variant, #444748);
--- a/src/pages/GlobalSettings/styles.module.scss
+++ b/src/pages/GlobalSettings/styles.module.scss
@@ -11,8 +11,8 @@
     min-width: 0;
 }
 
-/* ORISO-Admin#735: the document master data card has four field groups, so it takes the
-   full width of the grid rather than sharing a row. */
+// ORISO-Admin#735: the document master data card has four field groups, so it
+// takes the full width of the grid rather than sharing a row.
 .documentMasterDataCardSlot {
     min-width: 0;
     grid-column: 1 / -1;


kiodreambau:
This looks superseded by #784.

Both PRs fix the same two stylelint errors in the same two files — `src/components/Dpia/styles.module.scss` and `src/pages/GlobalSettings/styles.module.scss` — with different answers: #784 converts the comments to `/* … */`, this one to `//`. #784's own title already names the situation ("two stylelint errors, two different right answers"). This PR changes nothing else; the diff against `pre-dev` is those two files only.

With #784 applied, the gate is green: `npx stylelint "src/**/*.(s(c|a)ss|css|less)"` reports **0 errors** (1533 warnings, exit 0).

So one of the two should be closed. Which one is a style call — `//` is arguably the more idiomatic SCSS comment — but keeping both open means the next person hits the same conflict I did.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

kiodreambau:
Supersedes #784, now closed, which fixed the same three errors via `ignore: ["after-comment"]` in `stylelint.config.js`.

Recording the reason here so it is visible from this PR: `ignore` is evaluated before `except` in `rule-empty-line-before`, so that option also stops enforcing the existing `except: ["after-single-line-comment"]`. Measured on the same probe file — a blank line after a single-line comment:

| config | result |
|---|---|
| `pre-dev` today, and this PR (no config change) | `✖ Expected no empty line before rule` — exit 2 |
| #784 (`ignore: after-comment`) | clean — exit 0 |

Both branches were green in CI, so the relaxation would not have shown up at review time.

kiodreambau:
Superseded by #788, which landed the same three fixes on `pre-dev` in commit `8a78937 fix(admin): clear the three stylelint errors blocking every PR (#788)`.

Verified on `origin/pre-dev` today:

- `src/components/Dpia/styles.module.scss` — blank line before the amber-family comment is present; blank line before `.chipDisabled` is present.
- `src/pages/GlobalSettings/styles.module.scss` — blank line before `.documentMasterDataCardSlot` is present.
- `CI - Main` is green on `pre-dev` for `adfc9e0c`, `02359efe` and `771d5b5b`.

`#788` chose blank lines where this PR converted the block comments to `//`. Both clear `comment-empty-line-before` and `rule-empty-line-before`; the gate is green either way, so the remaining diff here is cosmetic and the branch is `CONFLICTING`. Closing rather than rebasing a duplicate.

Note for anyone whose PR still shows `Lint (Stylelint)` red: the branch is behind `pre-dev` and re-running against the current base clears it.

Issue #713, opened 2026-08-12
Tenant-invite composer: primary actions render off-screen and the visible action is disabled without explanation


Storypapst:
## Problem

On `/admin/links/tenants` the tenant-invite composer's primary actions are rendered **outside the visible area** and cannot be reached without horizontal scrolling. This was reproduced at 1440px and again at **1600px** width — both ordinary desktop sizes.

Visible at 1600px: `E-mail`, `First name`, `Name`, `Tenant ID (Auto)`, and two chevron buttons. Cut off on the right: the send/invite actions, plus the table's own `Link` and `Actions` columns.

The `Send directly` control that *is* reachable by accessible name is a disabled segment (`class="_segment_… _main_…"`, `disabled` attribute present) and it stays disabled after all three text fields are filled and blurred. There is **no validation message, no hint, and no indication of what is missing** — `.ant-form-item-explain` and `[role="alert"]` are both empty.

## Actual behavior

An admin filling in a valid tenant invite has no visible way to send it, and no feedback explaining why the one control they can see is greyed out.

## Expected behavior

- The composer's primary action is visible at standard desktop widths without horizontal scrolling.
- A disabled primary action states what is missing (project rule: disable rather than hide, but always explain).

## Evidence

Reproduced with Playwright against Pre-Dev as super-admin `snowball.two@dreambau.de`, admin build `oriso-admin@sha256:164ecfd8…` (current `pre-dev` at the time of writing).

```
enabled after fill: false
enabled after blur: false
aria-disabled: null   disabled: ""   class: _segment_15l3d_8 _main_15l3d_42
validation texts: []
```

Screenshot: `_e2e-artifacts/case-handover-20260812/screenshots/73-invite-send-disabled.png`

## Impact

Blocks canvas step 3 ("PA sends invite to tenant") of the legal-to-case-handover E2E run, OpenResilienceInitiative/ORISO-E2E#57. Note that invites *have* been sent successfully before — the table shows accepted tenants (`princess.pony@dreambau.de`, `ned.flanders@dreambau.de`, both `Accepted` / 2FA `ACTIVE` / Access `READY`) — so this is about the composer's current reachability, not a broken backend.

## Open question for whoever picks this up

Whether the disabled state is a genuine precondition (some field the composer needs that is itself off-screen, e.g. a required selection next to `Tenant ID`) or a regression. Determining that needs the actions to be visible first.

PR #818, opened 2026-08-18
style(editor): 16px horizontal inset for the chapter-chip bar

kiodreambau:
## Why

Owner report (2026-08-18): the chapter-chip bar in the legal-document reader needs more horizontal padding left and right, and had drifted upward relative to the design. Measured on `origin/pre-dev` (Playwright, computed styles): both read-mode bar variants had **0px horizontal padding** — `chip.left == text.left`, the chips sat flush with the text edge.

**Regression attribution for the navigation defect reported alongside this** (chips not navigating in read mode): that is *not* broken on `pre-dev`. It was caused by the read-only editor never stamping heading ids for published HTML without them, was repaired on `pre-dev` last night by `321bdd8` ("Repair legal document read mode", PR #813, closes #811), and the storage-side root cause — ORISO-AgencyService's sanitizer stripping heading ids on save (TenantService fixed the identical gap in `7f37f30` on 2026-07-04; never ported) — is fixed in ORISO-AgencyService PR #271. The Pre-Dev environment still shows the defect because its Admin deployment runs a demo-branch image (`speed-16aug`) that predates `321bdd8`; getting the fix in front of the user is a deploy/image-provenance decision, not a further Admin patch.

## What

16px horizontal inset for the chip row in both read-mode variants of `M3RichTextEditor.module.scss`:

- fluid reader (sticky bottom bar): `padding: 16px 0 8px` → `16px 16px 8px`
- deck card: `padding: 16px 0 0` → `16px 16px 0`

The bar's vertical position (16px top spacing per #811) is unchanged — `321bdd8` already set it. The base `.RichEditor-anchorNav` rule used by the authoring form editor (8px 12px) is untouched.

## Verification (real browser, Playwright, local Storybook — Pre-Dev untouched)

Measured after: nav padding `16px 16px 8px` / `16px 16px 0`, chips inset 16px from the text edge.

Navigation re-verified on the same build, before and after the style change, at 390 / 820 / 1440 on: `DpaLegalReader` (desktop-with-chapters), TenantOnboarding organisation+DPA long text, DPA blocker (mobile + fullscreen dialog), `M3RichTextEditor` read-only-with-anchors, `DataProcessingAgreementCard` read-only. Every chip click scrolled the document and moved focus onto the target heading; keyboard activation (Tab to chip, Enter and Space) does the same in the reader, the 390px wizard and the fullscreen dialog.

[...]

PR #819, opened 2026-08-18
Tenant-invite surface: preview what is actually sent, and put the toolbar back on one scrolling line

kiodreambau:
Continues the tenant-invite surface work from `0 - Docs/plans/2026-08-18-tenant-invite-before-state.md` (findings A1–A4, B1–B6, C1–C5, D1–D5, E1–E4). Three commits were inherited; five are new. **I re-verified the inherited claims by mutation rather than trusting their commit messages** — see the adversarial pass.

## What #751 got

**Nothing. Its files are deliberately untouched, and that is the point.**

#751 is approved and wires a *third* client-side `EmailKitPreview` into this composer — the opposite of finding E2. Rather than collide over `EmailTemplatesDialog.tsx`, which #751 rewrites, the backend render was put **inside `EmailKitPreview` itself**. That component is already on `pre-dev`; #751 only consumes it. So #751 inherits the server-rendered preview without a merge conflict and without me editing another author's PR. Same for D1: the `tokenPicker` opt-out lives in `PlaceholderTemplateFieldConfig`, in the module.

### Merge order — a request, not an assumption

I ran the real merge in a throwaway branch and ran the **full** suite:

```
merged branch:  32 failed | 1932 passed
minus 6 pre-existing failures (lazyNamed + LegalSettings; they fail identically
on my branch alone — localStorage/sessionStorage environment)
= 26 real merge-order breaks
```

20 of them are one line:

```
Error: [vitest] No "previewInviteEmailTemplateContent" export is defined on the
"../../api/accountInvites/accountInvites" mock. Did you forget to return it from "vi.mock"?
```

#751's `EmailTemplatesDialog.test.tsx` mocks the API module with a factory returning only `listInviteEmailTemplates`, `createInviteEmailTemplate`, `updateInviteEmailTemplate`. Once `EmailKitPreview` imports the preview call, every test in that file that mounts the editor throws. **Both PRs are green alone and break each other on whichever lands second.**

The remaining 6 (`AccountInvitesBulk`, `AccountInvitesTab`, `PlaceholderTemplateEditor`) came from my resolving the conflict naively as "theirs", which discards commit `8715f3f` and the raw-token change. That is not a defect in #751 — it is the checklist of hunks a real merge has to reconcile by hand.

**Please merge #751 first, then I rebase and carry the one-line mock fix here.** Either order needs the same line added; only that order gives it an owner. If yours lands first, the fix has to happen on #751's open branch, which I will not do. If Frank prefers the other order: merge this, then #751 rebases and adds the line.

---

### E2 — the preview showed a mail nobody would ever receive

The composer previewed a hand-written React re-implementation of the mail frame (`invitePreviewMarkup.renderInviteEmailPreviewHtml`) with its own brand constants, its own token substitution and no CTA, while the dispatcher rendered something else. Two documents by construction.

`EmailKitPreview` now renders what `POST /service/useradmin/invite-email-templates/preview` returns — the same `InviteEmailPreviewService` the send path runs. The endpoint already existed on `pre-dev`; this is wiring.

Deliberate consequences, each of which a reviewer will otherwise want to "fix":

- **The editor stopped substituting sample values client-side** and sends the authored text raw (`InviteEmailTemplateEditor`). Substituting in both places would have replaced one preview/send divergence with a subtler one: the author's substitution vs. the recipient's.
- **`sandbox="allow-same-origin"` without `allow-scripts`.** Scripts in the mail document never execute, while the parent can still measure it and fit the frame to its content — so auto-fit was not sacrificed. The attribute carries a comment saying that adding `allow-scripts` alongside it defeats the sandbox entirely. If you are here to make the preview's links clickable: no.
- **A failed render shows as a failed render, with a retry.** A silent local fallback would look plausible and lie about what gets sent — the exact defect being removed.
- **A late response for older input cannot overwrite a newer preview.** The effect cleanup cancels it; the test resolves the stale promise *after* the rerender to force the race.

`invitePreviewMarkup.ts` becomes test-only once #751 lands and should be deleted in that follow-up. Not deleted here, to keep the merge clean.

### The duplicate link in the sent mail

Fixed in ORISO-UserService#1058 — the layout owns the action link, the body must not carry it, enforced in the shared renderer. Rationale, the precisely-defined strip, and the "what does an author do instead" answer are in that PR.

The Admin half: `{{inviteLink}}` is gone from `INVITE_EMAIL_TOKENS`. **Deliberately no save block** — the renderer strips a stored token structurally, so an existing template can always still be saved. An author who edits a word in a legacy template must never hit a wall.

### A1 — the toolbar scrolls again, and #713 was wrong

A1 reverses ORISO-Admin#713, on evidence. #713 made the row wrap because at 1600px it demanded 1781px in 1336px and the send button sat at x=1718 — *"unreachable"*. That word does not survive measurement: `.scroller`, the row's direct parent, is a real scrollport (`overflow-x: auto`). Measured in Storybook with Playwright on the restored `max-content` row:

| Viewport | scrollWidth / clientWidth | scrollLeft reaches | send button after scrolling |
| --- | --- | --- | --- |
| 1440 | 1781 / 1440 | 341 | left 1197, right 1440 — visible |
| 820 | 1781 / 820 | 961 | visible |
| 390 | 1781 / 390 | 1391 | left 147, right 390 — visible |

The row overflows **and** scrolls, and the primary action is reachable everywhere. #713 read "overflows" as "is lost"; the fix it needed was a working scrollport, which it already had.

`globalSearchBar.styles.test.ts` encodes a layout *decision* as a test. It is **rewritten with these numbers and the reasoning**, not deleted — so the next reader sees a decision reversed on evidence rather than a rule quietly dropped.

### A2 — one baseline, as one number

Measured at 1440 before: search pill, both name fields and both split buttons 56px; the "⋮" more-button **48px**; the Träger-ID group **76px**, because `M3NumberField` always renders a supporting-text line under its 56px box. Centring three totals put their field boxes on three baselines.

After `align-items: flex-start` + more-button into the 56px band, at **all three viewports**: eight controls, box height **56** throughout, and `distinctBoxTops = [4]` — a single number, which is the line Frank drew across the toolbar.

### D3 — one surface, one colour

Before: open button segment `rgba(0, 0, 0, 0)` against menu sheet `rgb(246, 243, 243)`. After: both `rgb(246, 243, 243)`, `equal: true`. Spelled `.outlined.open .segment` because it must outrank `.outlined .segment`, which sets transparent.

### D1, D2, D4, E1

- **D1** — `PlaceholderTemplateFieldConfig` gains `tokenPicker`; the subject opts out. In the module, so #751 inherits it.
- **D2** — the Active toggle is the shared `MuiSwitchField`. **The test deliberately does not assert `role="switch"`**: antd and MUI both expose it, so that assertion would have passed against the very component being removed. It asserts whose markup it is — no `.ant-switch`, yes `.MuiSwitch-root`. Please don't simplify it back into a role check.
- **D4** — `Lisa`/`Beispiel` → `Maren`/`Muster`. **This finding moved repos:** with a server-rendered preview the visible names come from `InviteEmailPreviewService`, so the change had to land there to be visible at all. Had it been applied only to the Admin token samples, the ticket would have closed with nothing visibly changed. The Admin samples follow so the picker's hints agree.
- **E1** — the BETREFF caption takes the tenant primary. Asserted against the stylesheet, since CSS modules are proxied under vitest and a rendered colour cannot be read back.

---

## Coverage pass — every finding

| # | State | Evidence |
| --- | --- | --- |
| A1 | **Fixed** | measurements above; contract test rewritten; screenshots ×3 |
| A2 | **Fixed** | `distinctBoxTops = [4]`, all boxes 56px, ×3 viewports; style test |
| A3 | **Fixed** (inherited `8715f3f`) | proven by mutation, below |
| A4 | **Fixed** (inherited `9057495`) | proven by mutation, below |
| B1/B2 | **Baseline, not defects** | context screenshots |
| B3 | **Fixed** (inherited) | proven by mutation |
| B4 | **Fixed** (inherited) | proven by mutation |
| B5 | **Fixed** (inherited) | inert bulk-mode chevron removed (`InviteComposer.tsx:511`) |
| B6 | **Deferred — root cause found, not fixed** | see below |
| C1–C5 | **Fixed** (inherited `5365d48`) | C3 verified covering both places; proven by mutation |
| D1, D2, D3, D4 | **Fixed** | tests + measurements above |
| D5 | **Withdrawn** | nothing built, per Frank |
| E1 | **Fixed** | style test |
| E2 | **Fixed** | this PR |
| E3 | **Fixed as a side effect** | UserService#1058 |
| E4 | **Deferred (PRIO LOW)** | not investigated; no code written |

### B6, deferred with its root cause

„Unklar warum?" — a DRAFT row appears pre-checked. There is **no auto-select at mount** (`AccountInvitesTab.tsx:56` initialises `[]`; the only effect at `:400-404` prunes, never adds). Two real causes:

1. `AccountInvitesTab.tsx:495` — after a *partially failed* bulk send, `setSelectedIds(failed.map(...))` keeps rows checked across the following `loadInvites()`, so they read as pre-checked on the next visit.
2. `M3Checkbox` renders disabled boxes at `opacity: 0.38`; on a filtered view where only DRAFT/EMAIL_SENT rows are selectable, the single full-opacity box reads as "checked".

Both are real, neither is a one-line fix (one is a selection-lifecycle decision, the other a control-state design question), and guessing Frank's intent here would be worse than naming it. Deferred with the pointers rather than half-fixed.

---

## Adversarial pass — I did not trust the inherited commits

Each production hunk reverted, then the tests run:

| Claim | Mutation | Result |
| --- | --- | --- |
| bulk send uses `/send` for DRAFT (data loss) | `deliver` → always `resendAccountInvite` | **1 failed** — proven |
| A4 search | `matchesInviteQuery` → `return true` | **3 failed** (2 unit + 1 board) — proven |
| C3 tooltips | FilterChip tooltip text → `""` | **2 failed** — proven |
| A3/B3 button variant | tonal-when-not-ready restored | **2 failed** — proven |
| duplicate link (UserService) | `renderBody()` → `render()` | **4 failed** — proven |
| E2 | fix absent = the starting state | **2 failed / 7 passed** before, **9 passed** after — proven |
| D1, D2, D4, E1, token removal | tests written first | **6 failed**, then implemented — proven |

**C3 additionally verified by hand**, since the brief doubted it: the tooltip covers the filter chip (`InviteProgressBoard.tsx:242`) *and* the row badge (`:379-392`), all six statuses, de + en, via one key set `links.accountInvites.statusHint.*`. The claim holds.

**A1/A2/D3 proven in a browser, not only in a stylesheet** — the numbers above come from `getBoundingClientRect` / `getComputedStyle` via Playwright.

### Where the two passes disagree

**C4.** The coverage pass marks it fixed, and `5365d48`'s regression test is genuine: row actions are real `<button disabled>`, never anchors. But Frank's annotation was „Warum links haben im **inactive State**" — an appearance complaint, not a markup one. I verified the markup and did **not** verify the computed cursor/colour of a disabled action icon in a browser. If he still sees link-blue, the cause is a global stylesheet, not this cell. **Treat C4 as "markup correct, appearance unverified"** rather than closed.

---

## Verification

```
Admin:  npx vitest run   ->  281 passed | 4 failed (285 files)
                             1918 passed | 6 failed (1924 tests)
```
The 6 failures are **pre-existing and unrelated** (`lazyNamed`, `LegalSettings` — localStorage/sessionStorage environment). They fail identically on this branch with none of my changes applied; verified against the baseline before starting.

`npx tsc --noEmit` clean · `eslint --max-warnings=0` clean · `stylelint` 0 errors · `prettier --check` clean.

## Evidence

Screenshots: `~/ORISO/_e2e-artifacts/tenant-invite-2026-08-18/screenshots/`, indexed in `../INDEX.md`.

**Pre-Dev was not touched and not deployed to.** Every shot is from local Storybook. `before_` is `origin/pre-dev` @ `762673e`, checked out into the same worktree so the same dev server, story and viewport render both halves; `after_` is this branch. Every pair is byte-different.

| Finding | Pair | Viewports |
| --- | --- | --- |
| A1, A2 (+ B/C context) | `*_A1-A2-B-C_tenant-invites_*` | 390 / 820 / 1440 |
| D1, E2 | `*_D1-E2_invite-email-editor_1440` | 1440 |
| D3 | `*_D3_split-button-open_1440` | 1440 |

Two honest gaps, stated rather than papered over:

- **D2 and E1 have no screenshot pair.** They live in `EmailTemplatesDialog`'s create form, whose Storybook story renders the trigger only; the form did not open under Playwright at any viewport, and the resulting pairs were byte-identical. I deleted them rather than ship evidence of nothing. Both are covered by the tests named above.
- **E2's after-shot shows the preview's failure state**, because Storybook has no backend. That *is* the fix: the composer no longer fabricates a plausible mail locally. The `before_` half shows the locally-rendered mail this change removes.

Closes nothing yet — findings map to the plan document, not to issues.


kiodreambau:
## Rebased onto `pre-dev` — #751 integrated

Merged `origin/pre-dev` (now `fac48c0`, so #751 **and** #818). My changes land **on top of** #751's structure, not instead of it.

### Resolved by hand (6 conflicts)

| File | Resolution |
| --- | --- |
| `EmailKitPreview.tsx` | **Both sides kept.** My backend render + #751's language resolution — `safeLanguageTag(language, safeLanguageTag(i18n.language))` now decides the language the *renderer is asked for*, so an English template still previews as English. Both sides had independently chosen `sandbox="allow-same-origin"` for the same reason, so that needed no decision. |
| `InviteEmailTemplateEditor.tsx` | Kept #751's `onManageTemplates`, per-kind `tokens` and `language`; kept my raw pass-through and D1. `tokens` now drives the **pickers only**, not the preview. |
| `EmailTemplatesDialog.tsx` | **Theirs wholesale.** My `Form.useWatch` preview wiring is superseded — the editor owns the preview and reaches the backend through `EmailKitPreview`. That is precisely why the seam was put there. Re-applied on top: `kind={draftMeta.kind}` and D2. |
| `AccountInvitesTab.tsx` | Both sides kept (my `onSearchQueryChange` + their two template props). |
| `EmailKitPreview.test.tsx` | #751's five language cases **re-pointed, not deleted** (see below). |
| `EmailTemplatesDialog.test.tsx` | Theirs + the one-line mock + three tests re-pointed. |

### D2 needed a real change, not a re-apply

#751's Active toggle is a plain controlled antd `Switch` next to its own `<label htmlFor>` — **not** a `Form.Item`. `MuiSwitchField` binds through antd's Form and does not fit there. I extracted the bare presentational **`MuiSwitch`** from it, so call sites that own their own state have somewhere to go other than antd. D2 holds on the new structure.

### Tests re-pointed rather than deleted

Three of #751's tests asserted the composer's **own client-side substitution**, which E2 removes by design. They now assert the new boundary — what the composer hands the renderer, and what it does with the answer. Likewise #751's `<html lang="…">` cases now assert the language the renderer is asked for. **The poisoned-language test matters more than before**, not less: that value now reaches a backend, and is still rejected.

### One thing genuinely lost, named rather than hidden

#751's *"applies the tenant primary colour from the admin theme to the shell"* is gone. It asserted that the Admin's `--m3-primary` CSS variable was inlined into a locally built document. There is no local document any more — branding comes from the backend's `EmailBrandingResolver`, which is the *correct* source (it is what the recipient actually gets). No behaviour is lost; an Admin-side assertion about Admin-side branding is, because the thing it asserted about no longer exists here.

### Also fixed: the same omission in two more files

`AccountInvitesBulk.test.tsx` and `AccountInvitesTab.test.tsx` had the identical missing preview mock. Added — an unmocked call there is a load-dependent hang, not an honest failure.

### Verification — full suite, not just touched files

```
npx vitest run   ->   279 passed | 6 failed (285 files)
                     1975 passed | 10 failed (1985 tests)
```
**The 26 predicted merge breaks are gone.** The remaining 10 are pre-existing on `pre-dev` itself — I measured them there rather than assuming: a detached worktree at `origin/pre-dev` (`fac48c0`, none of my code) fails **the same 10**:

- `lazyNamed` ×1, `LegalSettings` ×5 — localStorage/sessionStorage environment
- `AccountInvitesTab` ×2, `AccountInvitesBulk` ×2 — 30s timeouts on `pre-dev`, unrelated to this branch

Worth flagging separately: **`pre-dev` is currently red on those four**, independent of this PR.

`tsc --noEmit` clean · `eslint --max-warnings=0` clean · `prettier --check` clean · `stylelint` 0 errors.

Not merging this myself, as agreed.

kiodreambau:
## Correction: I was wrong about `pre-dev` being red — retracting that

My previous comment claimed `pre-dev` is shipping red on four invite tests. **That claim was wrong and I am withdrawing it.** I was measuring this machine, not the branch. Both halves, re-measured properly:

### The full suite is green — 1985/1985

```
node v22.22.0 (CI uses 22.12.0)
npx vitest run  ->  285 passed (285 files) | 1985 passed (1985 tests)
```

Zero failures. Not "green except known issues" — green.

### Where my 10 failures actually came from

**Six were a Node version difference, not a defect.** CI pins `node_version: '22.12.0'` (`.github/workflows/ci-*.yml`); this machine runs **v26.5.0**, which ships an experimental global `localStorage`/`sessionStorage`. The `LegalSettings` and `lazyNamed` tests assert the storage-unavailable path, so that global inverts them. Proven rather than assumed — the identical checkout, only the interpreter swapped:

| Node | Result |
| --- | --- |
| v26.5.0 | 6 failed \| 281 passed |
| v22.22.0 | **287 passed (287)** |

**Four were load flakiness.** `AccountInvitesTab` ×2 and `AccountInvitesBulk` ×2 hit the 30s `testTimeout` while three agents were running on this box (load average 20.15 over the relevant 15 minutes, including a Testcontainers MariaDB). In isolation they pass — on my branch (21/21) *and* on pristine `origin/pre-dev` @ `fac48c0` (17/17).

The clue was in my own data and I read past it: a single *passing* test took **28.29s of a 30s budget**. I reported the conclusion I had reached instead of the one the number supported.

### What is worth keeping from this

Two small findings, neither blocking, neither about `pre-dev`'s health:

1. **Those four tests sit at ~28s against a 30s timeout.** That is one background job away from red, and it will bite CI eventually. Worth a look as its own task — not in this PR.
2. **The suite is not Node-version-portable.** Anyone developing on Node 23+ sees six red tests that CI never shows, which trains people to ignore a red suite. An `engines` field or an `.nvmrc` would make the expectation explicit.

Apologies for the noise — `pre-dev`'s test health is fine, and nothing here blocks this PR.

Issue #820, opened 2026-08-18
Test suite is not portable across Node versions and the expected version is undeclared


kiodreambau:
## Why

The Admin test suite is not portable across Node versions, and nothing in the repo declares which version is expected. Measured on 2026-08-18 on an identical checkout of `pre-dev`, swapping only the interpreter:

| Node | Result |
| --- | --- |
| v26.5.0 | **6 failed** \| 1979 passed |
| v22.23.2 | 1985 passed |

CI only ever runs the pinned `22.12.0`, so it never shows this. A developer on a current Node sees a red suite that CI calls green — and learns to ignore red. That is how a real failure gets waved through.

### Root cause

It is not "the tests rely on Web Storage being absent". Node 23+ ships an experimental global `Storage` class. Verified with a probe under Vitest + jsdom:

| | Node 22.23.2 | Node 26.5.0 |
| --- | --- | --- |
| `typeof globalThis.Storage` | `undefined` | `function` |
| `window.localStorage.constructor.name` | `Storage` (jsdom) | `MemoryStorage` (our polyfill) |
| `getPrototypeOf(window.localStorage) === Storage.prototype` | `true` | **`false`** |

Raw jsdom provides a real `Storage` on both versions, so jsdom is not the variable. Under Vitest `window === globalThis`, so on Node 26 Node's own web-storage globals share a namespace with jsdom's DOM globals; the `if (!window[name])` guard in `src/test/setup.ts` then installs our `MemoryStorage`, whose prototype is unrelated to the surviving `Storage`.

The six affected tests all simulate a full or disabled storage with `vi.spyOn(Storage.prototype, 'setItem' | 'removeItem' | 'getItem')`. On Node 26 that spy patches a prototype nothing ever calls, so the code under test quietly *succeeds* where it is expected to report failure.

Affected: `legalDraftStorage` (2), `useLegalDraft` (2), `LegalText` (1), `lazyNamed` (1).

## What

1. Declare the expected Node version once and make CI read it, instead of the version living only in workflow YAML.
2. Fix the tests so they survive a runtime that provides the storage globals, rather than depending on their absence.

## Goal

The suite produces the same result on any supported Node, and the expected version is discoverable from the repo itself.

**Acceptance criteria**

- [ ] `.nvmrc` exists and matches the version CI installs.
- [ ] `package.json` declares `engines.node`, consistent with `.nvmrc`.
[...]

PR #824, opened 2026-08-18
chore(ci): declare the Node version once and make the storage tests runtime-independent

kiodreambau:
## Problem

The test suite is not portable across Node versions and nothing in the repo declares the expected one. Identical checkout of `pre-dev`, only the interpreter swapped (2026-08-18):

| Node | `npm test` |
| --- | --- |
| v26.5.0 | **6 failed** \| 1979 passed |
| v22.23.2 | 1985 passed |

CI only runs the pinned `22.12.0`, so it never shows this. A developer on a current Node sees a red suite CI calls green, and learns to ignore red.

The cause is not "the tests need storage to be absent". Node 23+ ships a global `Storage` class; under Vitest `window === globalThis`, so the `if (!window[name])` guard in `setup.ts` installed a `MemoryStorage` whose prototype was *not* the surviving `Storage.prototype`. The six tests simulate a disabled storage via `vi.spyOn(Storage.prototype, …)` — on Node 26 that spy patched an object nothing called, so the code quietly succeeded where it must report failure.

## Changes

- Add `.nvmrc` (`22.12.0`) and `engines.node` to `package.json`.
- `node-build` action now reads `node-version-file: .nvmrc`; the `node_version` input and all four hardcoded call sites are gone — one source of truth.
- Install the in-memory storage **on `Storage.prototype`** so prototype spies intercept, for both areas unconditionally (presence ≠ usability).
- Add `memoryStorage.test.ts` guarding the spy-ability contract, so breaking it fails *here* instead of in an unrelated feature.

## Verification

- `npm test` on Node 22.23.2 — 286 files / 2010 tests, all green.
- `npm test` on Node 26.5.0 — the 6 storage failures are gone.
- `npx eslint . --max-warnings=0`, `npx prettier . --check`, `npm run build` — green.
- Verified local only; no deploy.

**Not fixed here:** `src/pages/Links/*` (`AccountInvites*`, `InviteComposer`, `EmailTemplatesDialog`) is flaky on Node 26 **before and after** this change — an untouched `pre-dev` run showed 8 files / 22 tests failing, and these pass in isolation. Order-dependent, out of scope. `Dockerfile.storybook` still carries its own `ARG NODE_VERSION=22.12.0`; a Docker `ARG` default cannot read `.nvmrc` without changing the shared image-build action, so I left it.

## How should I test this?

- [ ] `cat .nvmrc` → `22.12.0`; `node -p "require('./package.json').engines"` → `{ node: '^22.12.0' }`.
- [ ] `grep -rn "node_version" .github/` → no hits; the action reads `.nvmrc`.
- [ ] With Node 22 active, `npm test` → all green.
- [ ] With Node ≥23 active, `npx vitest run src/test/memoryStorage.test.ts src/pages/lazyNamed.test.ts src/components/Tenants/LegalSettings` → all green (was 6 failures).
- [ ] CI on this PR installs Node 22.12.0 in the "Set up Node.js" step log.

Closes OpenResilienceInitiative/ORISO-Admin#820


kiodreambau:
CI evidence — `setup-node` resolved the version from `.nvmrc`, no hardcoded input anywhere ([run 32129431148](https://github.com/OpenResilienceInitiative/ORISO-Admin/actions/runs/32129431148)):

```
node-version-file: .nvmrc
Attempting to download 22.12.0...
v22.12.0
```

`CI - Feature Branch` and `CI - Storybook Pull Request` are green.

### On the `src/pages/Links/*` flake

To be sure this PR did not cause it, I ran the full suite three times on an **untouched `pre-dev`** checkout with Node v26.5.0:

| baseline run | result |
| --- | --- |
| 1 | 4 files / **6** tests failed — exactly the storage six |
| 2 | 8 files / **22** tests failed |
| 3 | 7 files / **13** tests failed |

The storage six are deterministic; the `Links` family varies run to run and passes in isolation. It is pre-existing and order-dependent, not introduced here. Node 22 was 100% green in every run, before and after.

Shirloin (approved):

PR #832, opened 2026-08-18
fix: add team/single agency toggle to agency settings

Shirloin:
## Why
Admins had no way to view or change whether a counselling centre 
is a team or single agency. The field existed in the API but was 
never exposed in the UI.

Note: Issue [#1149](https://github.com/openResilienceInitiative/ORISO-Frontend/issues/1149) was filed against ORISO-Frontend but the fix 
belongs in ORISO-Admin — the agency settings form lives here.

## What
- Add MuiSwitchField for teamAgency in AgencySettings component
- Show warning when converting team agency to single 
  (existing team chats become 1:1)
- Add new i18n key agency.form.settings.teamAdviceCenter.changeWarning
- Form init: create defaults to single, edit keeps saved value
- 4 new tests: switch present, warning on team→single, no warning on create
- Fix: the `persistedTeamAgency` baseline the warning compares against is
  now passed in as a prop derived straight from the fetched agency data
  (`persistedTeamAgency={initialValues.teamAgency}`) instead of being
  captured once on mount into local state. Previously a second edit in the
  same page load (flip → save → flip again) compared against the stale
  pre-save baseline and could show the conversion warning incorrectly. New
  regression test: `refreshes the warning baseline after a save instead of
  comparing against the pre-save value`.

## Known issue — RESOLVED
Converting a team agency to single returned 500 — UserService threw 
LazyInitializationException in ConsultantAgencyAdminService.noOtherTeamAgency 
when accessing the consultantAgencies collection after the Hibernate session 
is closed.

Backend fix is up: **ORISO-UserService#1070** — https://github.com/OpenResilienceInitiative/ORISO-UserService/pull/1070, closing **ORISO-UserService#1069** — https://github.com/OpenResilienceInitiative/ORISO-UserService/issues/1069. Both directions of the toggle (team→single and single→team) work end to end once that PR merges.

## Test
- Toggle visible on /admin/agency/:id/general → Einstellungen tab ✅
- Save with changed toggle calls POST /changetype correctly ✅
- Warning shown when switching team → single ✅
- 4 unit tests passed ✅
- New: warning baseline correctly refreshes after a save (regression test) ✅

## Screenshot
image

Closes [#1149](https://github.com/openResilienceInitiative/ORISO-Frontend/issues/1149)

## Summary by CodeRabbit

* **New Features**
  * Added a team counseling setting for agencies.
  * Added a warning when disabling team counseling, explaining that existing team chats become 1:1 consultations and are no longer visible to other counselors.
  * Added English and German translations for the warning.
* **Bug Fixes**
  * Ensured the warning reflects the latest saved agency setting.
* **Tests**
  * Added coverage for switching, saving, and warning behavior.

### DIFF
--- a/scripts/en-translations-manual.json
+++ b/scripts/en-translations-manual.json
@@ -121,6 +121,7 @@
     "agency.form.registrationSettings.title": "Visibility in registration",
     "agency.form.registrationSettings.toPostCode": "To postal code",
     "agency.form.settings.headline": "Counseling service settings",
+    "agency.form.settings.teamAdviceCenter.changeWarning": "Turning off team counseling converts existing team chats into 1:1 counseling. Other counselors at this counseling center will no longer be able to view those chats.",
     "agency.form.settings.teamAdviceCenter.description": "All consultants of an advice center can view and participate in all chats with people seeking advice.",
     "agency.form.settings.teamAdviceCenter.title": "Offer team counseling",
     "agency.gender": "Gender",
--- a/src/components/CardEditable/CardEditable.test.tsx
+++ b/src/components/CardEditable/CardEditable.test.tsx
@@ -40,6 +40,23 @@ describe('CardEditable', () => {
         expect(screen.getByRole('switch', { name: 'Enabled' })).toBeEnabled();
     });
 
+    it('submits a toggled-off switch as false so card patches keep the new value', async () => {
+        const user = userEvent.setup();
+        const onSave = vi.fn();
+        render(
+            <CardEditable titleKey="card.title" onSave={onSave} initialValues={{ teamAgency: true }}>
+                <MuiSwitchField name="teamAgency" label="Team" />
+            </CardEditable>,
+        );
+
+        await user.click(screen.getByRole('button', { name: 'edit' }));
+        await user.click(screen.getByRole('switch', { name: 'Team' }));
+        await user.click(screen.getByRole('button', { name: 'card.edit.save' }));
+
+        await waitFor(() => expect(onSave).toHaveBeenCalled());
+        expect(onSave.mock.calls[0][0]).toEqual({ teamAgency: false });
+    });
+
     it('keeps entered values editable when the save callback reports an error', async () => {
         const user = userEvent.setup();
         const onSave = vi.fn();
--- a/src/locales/de/translation.json
+++ b/src/locales/de/translation.json
@@ -622,6 +622,7 @@
     "agency.form.settings.headline": "Einstellungen zum Beratungsangebot",
     "agency.form.settings.teamAdviceCenter.title": "Teamberatung anbieten",
     "agency.form.settings.teamAdviceCenter.description": "Alle Berater:innen einer Beratungsstelle können alle Chats mit Ratsuchenden einsehen und sich daran beteiligen.",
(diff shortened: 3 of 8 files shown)


shazia-k (changes requested):
Nice patch overall Riccardo — tests are thorough, the switch placement looks clean, and the DE/EN warning copy reads well. I like that you added the `CardEditable` regression for toggled-off booleans too; that's a subtle one to catch.

One thing I want to sort out before this goes in: the team → single direction returning 500. That's the more destructive of the two directions (existing team chats collapse to 1:1), and it's also probably the more common reason an admin would touch this toggle. If we ship the switch now, admins will use it, click Save, and see a generic error with no way to tell whether the conversion happened. Also I couldn't find the tracking issue for the UserService `LazyInitializationException` — could you drop the link into the PR body? My preference would be to either land this together with the UserService fix, or narrow it to a read-only display of the value until the backend fix is in (that still solves the "admins can't see whether an agency is team or single" complaint from #1149). Happy to be talked out of it if you have a reason to ship both directions live.

Two smaller things while I'm here:

**`persistedTeamAgency` baseline never refreshes.** It's set once via `useEffect` guarded on `null`, so after a successful save the baseline stays stuck on the pre-save value. If an admin does two edits in the same page load (flip, save, flip again), the warning banner will compute against the wrong baseline. Small fix — reset it when the form's initial value changes, or key it off `agencyData?.id`.

**Warning is one-directional.** Only team → single shows the warning, but single → team also expands who can see the chats (every counsellor at the agency instead of just the assigned one). Worth checking with product whether the reverse direction deserves a warning too — not a blocker for me.

Nit: worth double-checking with design that the switch belongs at the top of the settings fields (above the tenant selector), the screenshot looks OK but placement wasn't in the ticket.

Thanks!

Shirloin:
UserService bug tracked in https://github.com/OpenResilienceInitiative/ORISO-UserService/issues/1069 — LazyInitializationException in ConsultantAgencyAdminService.noOtherTeamAgency when converting team → single agency. This PR should land alongside the UserService fix.

Storypapst:
Update addressing your review:

**Backend fix landed.** Opened OpenResilienceInitiative/ORISO-UserService#1070, closing OpenResilienceInitiative/ORISO-UserService#1069 — the `LazyInitializationException` you flagged. Root cause: `ConsultantAdminFacade#changeAgencyType()` runs outside an open Hibernate session (it's intentionally not `@Transactional`, and `spring.jpa.open-in-view` is disabled), so `noOtherTeamAgency()` was touching the lazy `consultantAgencies` collection on an already-detached `Consultant`. Fixed by querying `ConsultantAgencyRepository` directly instead — same explicit-fetch pattern already used in `ChatService` for the same class of bug. Covered by a new integration test that reproduces the original 500 against the pre-fix code and passes against the fix. Requested your review there too.

**`persistedTeamAgency` baseline fix**, pushed to this branch (`e4bca429`): the baseline is now passed in as a `persistedTeamAgency` prop sourced directly from the fetched agency data (`initialValues.teamAgency`) instead of being captured once into local state on mount. Previously, since `AgencySettings` stays mounted across `CardEditable` save cycles, a second edit in the same page load (flip → save → flip again) compared the live field against the stale pre-save baseline and could show the team→single warning incorrectly even when nothing had actually changed since the last save. `useAgencyUpdate` already updates the query cache synchronously on save, so this prop — and the baseline — now refreshes on its own the moment a save completes. Added a regression test: `refreshes the warning baseline after a save instead of comparing against the pre-save value`.

I did not touch the single→team warning — leaving that as product's call per your note.

Both directions of the toggle (team↔single) should now work end to end once ORISO-UserService#1070 merges. PR body updated with the link.

PR #833, opened 2026-08-18
fix(branding): show the uploaded favicon in the admin tab for everyone

kiodreambau:
## Problem

The platform admin's uploaded favicon never reached the browser tab for the
visitors it's meant for. Reported behaviour: "what the platform admin uploads
should be the default icon everyone — including unregistered visitors — sees."

## Root cause (both independent, both fixed here)

1. **Unreachable on anonymous routes.** The favicon-apply effect lived inside
   `<App />`, which sits behind `<ProtectedRoute>`. Every route an unregistered
   visitor actually renders — `/admin/login`, password reset, tenant/counsellor
   onboarding, imprint, privacy — never mounts `<App />`, so the effect never ran
   and the tab kept the built-in placeholder (`/admin/favicon.ico`).
2. **Only one of five icon links updated.** Even when the effect did run, it
   rewrote only `document.querySelector("link[rel='icon']")` — the first of five
   `rel="icon"` links `index.html` declares. Browsers pick the icon whose `sizes`
   best matches the tab (typically `favicon-32x32.png`), so the four sized PNG
   variants (192/96/32/16) kept showing the placeholder even when the first link
   was correctly rewritten.

Verified independently in this review, not just asserted by the commit message
(see "Fallback order" below).

## Fix

- Extracted the route composition from `index.tsx` into a new `AdminApp.tsx`
  (testable mount point; `index.tsx` keeps only module-level side effects).
- `<TenantFavicon />` now mounts **above `<Routes>`** in `AdminApp`, so it applies
  on every route, with or without a session.
- New `applyBrandingFavicon()` rewrites **every** `link[rel~="icon"]` (not just
  the first) and strips the `sizes`/`type` hints that described the placeholder
  PNGs, so no stale size hint can make a browser pick a wrong variant.
- New `resolveBrandingFavicon()` encodes the fallback chain (see below), reusing
  the existing `getSafeFaviconUrl` guard so an undecoded/broken candidate (e.g.
  leftover `&#61;` HTML entities) degrades to the next step instead of applying a
  broken icon.
- `<App />` now passes the signed-in admin's own tenant favicon into the same
  `<TenantFavicon tenantFavicon=... />`, so a tenant-specific icon still overrides
  the platform default once a session's own tenant data has loaded.

## Fallback order implemented

`resolveBrandingFavicon`: **tenant icon → platform icon → built-in**.

"Platform" is not a new concept introduced by this PR — it is the existing
single-domain-multitenancy "main tenant"
(`mainTenantSubdomainForSingleDomainMultitenancy`, a platform-global setting
already surfaced through `usePublicTenantData`/`getPublicTenantData`, which every
route already resolves via `GET /service/tenant/public/{mainTenantSubdomain}`,
`skipAuth: true`). I independently traced this through `ORISO-TenantService`
(`TenantController` → `TenantServiceFacade#findTenantBySubdomain` →
`SingleDomainTenantOverrideService`) to confirm `theming.favicon` for that
endpoint is always the main tenant's own row — `SingleDomainTenantOverrideService`
only ever overrides privacy content and a fixed list of media feature toggles,
never `theming`. So "the platform admin uploads a favicon for the main tenant,
and it becomes everyone's default" was already fully wired end-to-end on the
backend; **no TenantService/backend contract change was needed or made**. This
mirrors the precedent in `ORISO-UserService`'s `EmailBrandingResolver`
(tenant → platform → built-in for outgoing-mail branding), though that resolver's
"platform" step is a static Spring config value, not a real uploaded tenant — the
favicon case is arguably a cleaner instance of the pattern since the platform
default here really is admin-editable through the existing tenant-theming UI.

## Mutation-test result

Reverted each production hunk in isolation and confirmed the paired test(s) go
red, then restored (verified `git status`/`git diff` clean afterwards):

- `applyBrandingFavicon` reverted to "only rewrite the first icon link" (the old
  bug) → 5 tests failed across `applyBrandingFavicon.test.ts`,
  `TenantFavicon.test.tsx`, `AdminApp.test.tsx`.
- `resolveBrandingFavicon` reverted to drop the platform fallback (tenant-only) →
  4 tests failed across `resolveBrandingFavicon.test.ts`, `TenantFavicon.test.tsx`.
- `AdminApp`'s `<TenantFavicon />` moved back inside the protected route (the old
  mount location) → `AdminApp.test.tsx`'s anonymous-route test failed red exactly
  as expected.

All three mutations were confirmed red, then reverted back to the committed code.

## Evidence

Live pre-dev/dev network access was flaky for this session (known DNS/CA issue on
the shared pre-dev host), so before/after evidence was captured against a
throwaway local stub standing in for TenantService/ConsultingTypeService (same
code path: `usePublicTenantData` → `getPublicTenantData` → `TenantFavicon` →
`resolveBrandingFavicon` → `applyBrandingFavicon`), on the anonymous
`/admin/login` route:

- **Before** (parent commit `3c8dad6`): all 5 `link[rel~="icon"]` hrefs stay on
  the built-in placeholder despite the stub serving a valid platform favicon.
- **After** (`840ba576`): all 5 hrefs are rewritten to the platform favicon data
  URL, `sizes`/`type` stripped.

Full before/after DOM capture kept locally (not attached to this PR body):
`/private/tmp/claude-501/-Users-kio-Documents-GitHub-ORISO/679645fb-2db9-4ce8-83af-101e19dd4cd0/scratchpad/favicon-evidence-dom-capture.md`.

`dreambau-evidence doctor` was run to publish this properly and failed on the
Keychain step specifically (not gateway/GitHub):

```
ok   git repository       840ba57
ok   gh authentication    signed in as kiodreambau
fail keychain token       invalid machine identity name
ok   gateway              https://evidence.dreambau.com/api/v1 — reachable
```

Reporting this plainly rather than faking an attachment: the evidence stays local
until the machine-identity Keychain entry is fixed.

## Reviewer test plan

- [ ] `npm run test` — full suite green (2042/2042 on Node 22.12.0 per `.nvmrc`)
- [ ] `npx eslint . --max-warnings=0` and `npx prettier . --check` on the changed files
- [ ] `npx tsc --noEmit` clean
- [ ] Visit `/admin/login` (or any anonymous route) with a tenant configured as
      `mainTenantSubdomainForSingleDomainMultitenancy` that has an uploaded
      favicon; confirm the tab icon is the platform favicon, not the placeholder
- [ ] Log in as a tenant admin whose own tenant has a different favicon; confirm
      the tenant's own icon overrides the platform default
- [ ] Inspect `document.head` and confirm all 5 `link[rel~="icon"]` hrefs match
      (not just the first) with `sizes`/`type` stripped


shanzaeimran (approved):

kiodreambau:
Follow-up opened as **#861**.

This PR merged before the last commit of the work could be pushed to it, and the head branch was deleted. The remaining commit (`c11732d2`) is the second half of the same fix and is now on #861:

`<TenantFavicon tenantFavicon={data?.theming?.favicon} />` is fed from `useTenantData()` → `getTenantData`, and that module — unlike `getPublicTenantData` — never ran `decodeTenantAsset`. TenantService stores `+` as `&#43;` and `=` as `&#61;`, `getSafeFaviconUrl`'s HTML_ENTITY guard rejects any data URL still carrying them, and `resolveBrandingFavicon` then silently fell through to the platform default. So the tenant override this PR added could never fire on the authenticated route.

#861 moves the decode into one place (`decodeTenantBrandingAssets`) used by both callers, so the two seams cannot drift apart again.

PR #839, opened 2026-08-18
fix(login): centre the password-reset link and use the M3 filled submit button

kiodreambau:
## Problem

Owner review of the Pre-Dev sign-in form: the `Forgot password?` link sits ~100px left of the axis the fields and the submit button share, and the submit button is an off-palette navy pill that reads as a stuck state layer rather than a solid M3 filled button.

## Root cause

The submit action was a raw MUI `<Button variant="contained">`. `theme/orisoMuiTheme.ts` gives MUI a single hardcoded `palette.primary.main = '#273270'` — the legacy antd `colorPrimary` that `theme/antdM3Theme.ts` retired everywhere else — so MUI derived the whole control from that literal: `--variant-containedBg: #273270`, elevation-2 **at rest** (raised to elevation-4 on hover, 6 on focus, 8 on active), no focus ring at all, and `rgba(0,0,0,.12)` / `rgba(0,0,0,.26)` when disabled. Nothing in the markup is a gradient — `background-image: none`, no `::before`/`::after`, no children — the "lighter middle section" is the white label plus that resting drop shadow on a button that already looks hovered. The same `#273270` literal also coloured the link.

## What changed

- `.forgotPW` is a block-level box with auto inline margins → on the form axis, hit target still the size of the label.
- Submit is the shared `M3Button variant="filled" block` — the same control `TenantOnboarding/OrganisationDpaStep` submits with. Colour from `--m3-primary` / `--m3-on-primary`, flat at rest, `color-mix` hover state layer, 2px `:focus-visible` ring, `loading` replaces the manual `CircularProgress` `startIcon`.
- The link takes `--m3-primary` too, so link and button stay on one palette.

## Verification

Real Chromium, `/admin/login`, 390x844 · 820x1180 · 1440x900, light and dark (the admin defines no `prefers-color-scheme` rules, so both render identically — measurements are byte-identical across schemes).

| measurement | before | after |
| --- | --- | --- |
| link centre − form centre | −119.57 / −100.82 / −100.82 px | **−0.01 px** at all three widths |
| resting background | `rgb(39,50,112)` (`#273270`, no token) | `rgb(165,0,10)` = resolved `--m3-primary` |
| resting box-shadow | MUI elevation-2 (3 stacked shadows) | `none` |
| hover | `rgb(27,35,78)` (MUI darken) | `color-mix(--m3-primary 90%, #000)` |
| `:focus-visible` outline | `0px none` (shadow only) | `2px solid rgb(165,0,10)` |
| height | 42.25px | 40px (M3) |

Screenshots (page + tight strips for disabled / resting / hover / focus, all three widths, no credential ever visible — the password field stays masked and the run aborts if `type !== "password"`) are attached below.

### Mutation test

Both hunks reverted individually against the new tests:

- revert `styles/components/loginForm.less` → `centres the password-reset link on the axis the fields and the submit button share` fails with `expected [ 'inline', 'inline-block' ] to not include 'inline-block'`; 12 other tests stay green.
- revert `pages/Login/LoginForm.tsx` → `renders sign in as the shared M3 filled button…` and `keeps the full-width submit block and the busy state…` fail on the missing `_filled_` / `_block_` class; 11 other tests stay green.
- both restored → 18/18 green in `src/pages/Login/`.

The centring test compiles the stylesheet the app really ships (`less.render` over `variables/index.less` + `loginForm.less`), applies it to the really rendered form and asserts through the actual cascade — not a regex over the source line.

### Commands

- `npx vitest run src/pages/Login/` → 18 passed
- `npx tsc --noEmit` → clean
- `npx eslint …` → clean · `npx prettier --check …` → clean · `npx stylelint …` → 0 errors
- `npm run test` (full) → the only failures are pre-existing: `src/pages/users/Edit/index.test.tsx` fails identically with this branch's files reverted to `origin/pre-dev`, and `src/pages/Links`, `DocumentMasterDataCard` time out non-deterministically under parallel load and pass in isolation.

## Follow-up, not in this PR

`M3Button`'s disabled state is `opacity: .38` on the fill, so a disabled **filled** button is a washed-out pink pill instead of M3's neutral `on-surface` 12%/38% treatment. That is existing shared-component behaviour (it already ships on the onboarding steps), and the sign-in form opens in exactly that state — but changing it repaints every disabled `M3Button` in the admin, so it wants its own decision rather than a drive-by edit here.

### DIFF
--- a/src/pages/Login/LoginForm.test.tsx
+++ b/src/pages/Login/LoginForm.test.tsx
@@ -1,8 +1,12 @@
 import React from 'react';
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+import less from 'less';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import LoginForm from './LoginForm';
+import m3ButtonStyles from '../../components/M3Button/styles.module.scss';
 import { FETCH_ERRORS } from '../../api/fetchData';
 import { ADMIN_PORTAL_ACCESS_DENIED } from '../../hooks/useLoginMutation.hook';
 import { TwoFactorType } from '../../enums/TwoFactorType';
@@ -34,6 +38,28 @@ const translations: Record<string, string> = {
 
 const t = (key: string) => translations[key] || key;
 let consoleWarnSpy: ReturnType<typeof vi.spyOn>;
+
+const STYLES_ROOT = resolve(__dirname, '../../styles');
+
+/** Compile the real login stylesheet (its LESS variables included) to CSS. */
+const compileLoginFormCss = async () => {
+    const entry = readFileSync(resolve(STYLES_ROOT, 'components/loginForm.less'), 'utf8');
+    const { css } = await less.render(`@import 'variables/index.less';\n${entry}`, {
+        paths: [STYLES_ROOT, resolve(STYLES_ROOT, 'components')],
+        filename: resolve(STYLES_ROOT, 'components/loginForm.less'),
+        javascriptEnabled: true,
+    });
+
+    return css;
+};
+
+const applyCompiledLoginFormStyles = (css: string) => {
+    const styleElement = document.createElement('style');
+    styleElement.textContent = css;
+    document.head.appendChild(styleElement);
+
+    return () => styleElement.remove();
+};
 
 vi.mock('react-i18next', () => ({
     useTranslation: () => Object.assign([t], { t, i18n: { language: 'en' } }),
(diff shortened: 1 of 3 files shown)


kiodreambau:
Visual evidence is captured but **not published**: this machine has no Dreambau evidence identity provisioned (`dreambau-evidence doctor` → `fail keychain token`, `~/.config/dreambau-evidence/identities/` does not exist), so there is no stable HTTPS link to bind to this PR. The gateway itself is reachable; only the token is missing, and it has to be provisioned rather than typed in by an agent.

Until then the run is on the Mac Mini at `0 - Docs/_artifacts/admin-login-pr839/` — `before/` and `after/` at 390x844, 820x1180 and 1440x900, each with the full page plus a tight strip of the link and the submit button in its disabled, resting, hover and keyboard-focus states, light and dark. `measure-before.json` / `measure-after.json` hold the geometry and computed-style numbers quoted in the description. No credential appears in any frame; the capture aborts if the password field is not `type="password"`.

Issue #842, opened 2026-08-18
DPA forward: the recipient name is collected but never sent, so the mail preview promises a salutation the backend cannot deliver


kiodreambau:
Found while reworking the forward dialog layout (#723/#724, epic #722).

## Problem

`DpaForwardRequest` in `src/api/tenantOnboarding/dpaForward.ts` carries **only** `recipientEmail`:

```ts
export interface DpaForwardRequest {
    recipientEmail?: string;
}
```

`DpaForwardDialog.submitEmail` accordingly calls `forward({ recipientEmail })` and drops `recipientName` on the floor. The field exists in the form and it drives the preview, but its value never reaches the backend — so the mail the recipient actually receives cannot contain the name the admin typed, while the preview shows it resolved.

The preview is therefore accurate for the empty case (a neutral salutation is what the backend can send) and **inaccurate as soon as a name is typed**.

## Options

1. Extend the forward contract (ORISO-TenantService#191) with an optional `recipientName` and pass it through — the preview becomes true again and the field earns its place.
2. Drop the name field and always preview the neutral salutation — smaller, honest, but loses a nicety the owner asked for.

Option 1 is the better product answer; it needs a backend change, so it is an owner/service call rather than an Admin-only fix.

## Not blocking

The empty-state half is already fixed in the layout PR: an empty name now yields the neutral salutation instead of a raw `{{recipientName}}`.

Storypapst:
I had my coding agent check this. @nikunjdecyb Admin #895 now sends the trimmed recipientName and explicitly includes this issue. Please verify the receiving TenantService contract and the actual delivered email: a named recipient must receive the same salutation shown in the preview, and an empty name a neutral one. Record that readback before closing the preview-versus-delivery mismatch.

PR #760, opened 2026-08-14
DPA forward-to-authorised-signer: wizard forward dialog with mail preview and friendly pending-signature dialog (#723, #724)

Storypapst:
Closes #723, closes #724 (epic #722).

## Problem

An admin who is not authorised to sign the DPA cannot finish onboarding: wizard step 1 only supports signing on the spot, and after login the `DpaBlocker` locks the whole admin app with no recovery path.

## What changed

- Shared `DpaForwardDialog`: copyable single-use sign link (with success feedback), optional e-mail send with recipient name/address, and the actual `DPA_FORWARD` mail rendered through the e-mail kit preview; note that the link stays valid until the contract is signed regardless of the sharing channel.
- Wizard step 1 (#723): secondary path "I am not authorised to sign" below the consent act; after confirming, the step flips to a calm on-hold state (no signer fields, no consent) and registration proceeds with `dpa.accepted: false`; completion mentions the notification mail.
- Legal Settings forward UI now opens the same dialog and gains the preview.
- Post-login (#724): with `forwardPending` set, the hard blocker is replaced by a friendly recurring dialog (copyable link, re-send via the shared dialog, "Later" dismiss). Fail-closed: without the flag the strict #572 gate is unchanged.
- **Contract alignment done** against ORISO-TenantService#191 and ORISO-UserService#1031 (both green), including the 2026-08-15 correction: single `POST …/onboarding/dpa-forward` returning `{signUrl, expiresAt}`; typed 400/404/409 errors and body-first 410 link death; **502 keeps the created link** and shows a "mail not sent" warning rather than a failure; the waiting state is the **additive boolean** `forwardPending` / `dpaForwardPending` — there is no `PENDING_FORWARDED` enum value anywhere in the client, and the invented client-side `forwarded: true` registration flag is gone (the server authorises `accepted: false` against its own `dpa_forwarded_at`).
- i18n de/en; vitest for dialog, 502 path, flow transitions, gate decision and gate rendering; Storybook stories for dialog, on-hold step, pending dialog.

## Verification

- `npm run test` — 251 files / 1614 tests pass
- `npx tsc --noEmit`, `npx eslint . --max-warnings=0`, `npx prettier . --check`, `npm run build` — all clean
- Screenshots (desktop + 320px) in the evidence comment.

## Reviewer test plan

[...]

PR #843, opened 2026-08-18
DPA forward dialog: mail-first layout, viewport gutter, wide two-column fields, prominent send action

### LINKED ISSUE #841: DPA forward dialog: mail-first layout, viewport gutter, wide two-column fields, prominent send action
Owner review of the forward dialog ("Forward to an authorised signatory", #723/#724, epic #722), 2026-08-18.

## Problem

1. The sign-link block sits **above** the mail preview, although the mail is the worked example of what gets sent and the link is the alternative to it.
2. The sheet grows past the top and bottom of the viewport — every other dialog in the product keeps a gutter.
3. It is too narrow on desktop, so `Name of the person (optional)` and `E-mail address` stack on every width instead of sharing a row down to phone size.
4. `Send e-mail` is a small outlined button on the left edge; it should read as prominent as `Copy link` and sit on the trailing edge.
5. The preview renders a literal `{{recipientName}}` while the name field is empty and only resolves once the user types (previously logged as J4). A raw template token must never be shown to a person.

## Goal

The dialog reads mail-first, keeps 24px of air above and below at any viewport height, uses one field row on desktop and stacks on mobile, gives its send action filled emphasis on a right-aligned row, and greets neutrally instead of leaking the template.

Reuse the existing house pieces (`FieldGrid`, `M3Button` variants, the `calc(100dvh - 48px)` gutter the DpaBlocker card and the onboarding sheet already use) — no new spacing values or one-off colours.

kiodreambau:
Closes #841. Follow-up filed as #842.

## Base branch — and why this PR is blocked

`DpaForwardDialog` exists only on the unmerged forward work, not on `pre-dev`, so branching from `pre-dev` would have meant no component at all. This PR is therefore **stacked on `feat/avv-forward-wizard-723` (PR #760)** and cannot merge before #760 does. That PR was `CONFLICTING`; I merged `pre-dev` into it and resolved the single conflict (the antd import line in `DataProcessingAgreementContainer` — `pre-dev` added `Check`, the branch dropped `Input`; both hold). **#760 is `MERGEABLE` again**, and this PR is blocked only on it being reviewed and merged.

## Problem

The dialog put the sign link above the mail it is an alternative to, grew past the top and bottom of the viewport, stacked name and e-mail at every width, offered its send action as a small outlined button on the left, and printed a literal `{{recipientName}}` into the preview while the name field was empty.

## What changed

- **Order** — the mail section (fields, send, preview) comes first; the sign link and its validity note sit below it, behind the divider.
- **Gutter** — the sheet is bounded to `calc(100dvh - 48px)` and the modal body scrolls, so the dialog keeps 24px of air top and bottom at any height. That is the gutter `DpaBlocker` and the onboarding sheet already use; bounding the *sheet* rather than the content keeps it honest when the title wraps on a phone. `Modal` gained a `className` prop for it.
- **Width** — 720 → 880, and the recipient fields moved into the shared `FieldGrid` (max 2 columns, 240px floor): one row from ~700px up, stacked on a phone. No media queries, no new spacing values.
- **Send action** — M3 `filled` variant on a trailing-edge row.
- **Preview** — `buildForwardMailPreview` composes the body from a separate salutation key, so an empty name greets neutrally instead of leaking the token, and a link that does not exist yet is named instead of printed as `{{link}}`.

## On the `{{recipientName}}` empty state

Shipping a raw template token to a person is a bug, not a placeholder convention, so the empty state now greets neutrally (`Guten Tag,` / `Hello,`). Appending an empty name to the existing greeting would only have swapped the token for `Guten Tag ,`, so the salutation is its own i18n key. The neutral wording keeps the existing register rather than switching to `Sehr geehrte Damen und Herren,` — happy to change that if you prefer the formal form.

Separately, and worth knowing: `recipientName` is **never sent** — `DpaForwardRequest` carries only `recipientEmail`. The preview's resolved salutation is a promise the current contract cannot keep. Filed as #842.

## TDD + mutation evidence

Tests written first, watched fail, then implemented. Every production hunk was reverted in turn and the suite re-run:

| Mutation | Result |
| --- | --- |
| link block back above the mail | RED |
| drop the `calc(100dvh - 48px)` bound | RED |
| drop `overscroll-behavior: contain` | RED |
| width back to 720 | RED |
| fields back in a plain `div` instead of `FieldGrid` | RED |
| send button back to `outlined` | RED |
| action row back to `justify-content: flex-start` | RED |
| raw `{{recipientName}}` for the empty name | RED |
| naive empty-name fallback (`Guten Tag ,`) | RED |
| raw `{{link}}` before the link exists | RED |

jsdom has no layout engine, so the two rules only an engine can execute (the viewport bound, the trailing-edge row) are asserted against the shipped stylesheet; everything a DOM can answer is asserted on the render.

## Verification

Real Chromium at 390×844 / 820×1180 / 1440×900, measured not eyeballed:

| | 390 | 820 | 1440 |
| --- | --- | --- | --- |
| gap above / below sheet | 24 / 24 | 24 / 24 | 24 / 24 |
| body scrolls (content overflows) | yes | yes | yes |
| preview before link block | yes | yes | yes |
| name + e-mail share a row | no (stacked) | yes | yes |
| send button background | `#a5000a` | `#a5000a` | `#a5000a` |
| send row `justify-content` / right-flush | `flex-end` / 0px | `flex-end` / 0px | `flex-end` / 0px |
| raw `{{token}}` in preview | none | none | none |

Also run: `npm run test` for the touched areas (26 files, 308 tests green — `DpaForwardDialog`, `Modal`, `LegalSettings`), `npm run typecheck:storybook`, `tsc --noEmit`, `eslint --max-warnings=0`, `stylelint` (0 errors), `prettier --check`, on Node 22.12.0. The `src/pages/Links/*` and `src/pages/users/Edit` suites time out under parallel load on this machine; they pass in isolation on this branch and on the base alike, so they are the known order/timeout flakes, not this change.


kiodreambau:
### Visual evidence — not published

The three-viewport screenshots were captured from this commit in real Chromium (390×844 / 820×1180 / 1440×900, plus scrolled-to-link and named-recipient states) and the measurements from them are in the PR body. They are **not** attached, because `dreambau-evidence doctor` fails on this machine:

```
ok   git repository
ok   gh authentication    signed in as kiodreambau
fail keychain token       invalid machine identity name
ok   gateway              https://evidence.dreambau.com/api/v1 — reachable
```

There is no `dreambau-evidence` item in the login Keychain and no `EVIDENCE_IDENTITY` configured here, so the gateway identity is missing rather than broken. Publishing the images another way would bypass the project's evidence rules, so the images stay local until the identity is provisioned. Reviewers can reproduce them from `Organisms/DpaForwardDialog` → `Default` / `Mobile` / `NamedRecipient` in Storybook.

Issue #605, opened 2026-07-30
Restore intrinsic desktop width for M3 legal editors


Storypapst:
## Why

PR #582 correctly made off-screen cards discoverable, added scroll-position feedback, and retained mobile stacking, but it also capped every legal card at 425px. That contradicts the intended desktop contract for the M3 document editors: DPA, imprint, and privacy need to retain their larger intrinsic/minimum reading and editing width. Compact non-editor cards should remain compact.

Parent: #568

## What to build

- Keep the CardDeck dots, arrows, scrolling, narrow-layout stacking, and generic width safety from #582.
- Apply a desktop-only minimum/intrinsic width to M3 document-editor items in LegalSettings.
- Do not broaden other settings cards or change mobile stacking.
- Add regression coverage for the distinct compact-card and document-editor width contracts.

## Acceptance

- [ ] On desktop, DPA, imprint, and privacy editor cards use their larger document-editor minimum width instead of the 425px compact-card cap.
- [ ] Compact legal settings cards remain 425px wide.
- [ ] Position dots, arrows, horizontal scrolling, and narrow/mobile vertical stacking remain unchanged.
- [ ] A regression test fails with the all-cards-425px implementation and passes with the corrected scoped width contract.
- [ ] Relevant tests, lint, formatting, and build pass.

PR #863, opened 2026-08-20
fix(agency): give the three legal editors the platform-admin width

kiodreambau:
## Problem

On the agency **"Rechtliches"** tab, the DPA, imprint and privacy cards were far too narrow to edit a legal document in. The very same M3 document editor is 800px wide on the platform admin's legal deck, but on the agency tab it was 392px — 342px of usable text column.

## Root cause

The three cards were rendered as plain `CardDeck` items, so they inherited the **deck default of 392px**. The platform admin's legal deck does not hit this because it scopes the #605 panel-card raster to its own document items; the agency tab never opted in.

## Fix

Widen the three agency cards to the same panel-card raster as the platform admin, instead of stacking them. The deck is a horizontal scroller, so all three stay side by side and stay reachable through the deck controls.

Measured at 1440x900, agency card / usable editor width:

| | card | usable editor |
|---|---|---|
| before | 392 | 342 |
| after | **800** | **750** |

Identical to the platform admin.

### Judgement call carried over from the commit

Unlike the tenant deck, this deliberately sets **no `--card-deck-item-min-width`**. That minimum is exactly what pushes the platform admin's own page into a horizontal scroll at tablet widths (measured: document scroll width 856 in an 820 viewport). The deck's built-in `min(320px, max-width)` floor is enough, and the max-width is capped by the same viewport raster `CardDeck` already uses, so the card can never outgrow the page area.

## Tests

- `src/pages/Agency/Edit/styles.test.ts` — new style assertions on the panel-card raster (24 added lines).
- Verified free of page-level horizontal scroll at **390x844, 820x1180 and 1440x900**.

## Reviewer test plan

- [ ] Open an agency, go to the **Rechtliches** tab. All three legal cards are ~800px wide and sit side by side in the deck.
- [ ] Compare against the platform admin legal deck — the editor text column should be the same width (750px) on both.
- [ ] Scroll the deck horizontally with the deck controls; all three cards remain reachable.
- [ ] At **1440x900** confirm the page itself never scrolls horizontally.
- [ ] At **820x1180 (tablet)** confirm the page itself never scrolls horizontally — this is the regression the missing min-width avoids.
- [ ] **Mobile 390x844**: the card falls back to the deck floor, is still usable, and the page does not scroll sideways.


kiodreambau:
## Evidence — measured, not photographed

**There is no screenshot for this PR, and here is why.** The change is on the agency "Rechtliches" **page** (`src/pages/Agency/Edit/index.tsx`). The repo has stories for the `CardDeck` molecule and for the individual legal cards, but **no story renders that page**, so Storybook cannot show it. Photographing it needs the admin app running and logged in against a live API, which was out of budget for this pass — and would not have produced a more trustworthy number than the measurement below.

Instead the mechanism was measured directly in Chromium against the **real `CardDeck` component** (story `molecules-carddeck--multiple-cards-per-item`), comparing the deck default against **exactly the rule this PR adds**:

| viewport | deck default | with the JOB1 rule | page scrolls sideways? |
|---|---|---|---|
| 1440 x 900 | 392 px | **800 px** | no |
| 820 x 1180 | 392 px | **524 px** (capped) | no |
| 390 x 844 | 358 px | 358 px (rule is ≥768px only) | no |

This reproduces all three claims in the commit message independently:

1. **392 → 800 px** at 1440x900, matching the platform admin's panel card.
2. At **tablet** width the `max-width` cap pulls the card down to 524px instead of letting it overflow. This is the judgement call in the description — the tenant deck sets `--card-deck-item-min-width` and pays for it with a page-level horizontal scroll at this width; omitting it here is what keeps the page clean.
3. **No page-level horizontal scroll at any of the three widths** (`document.scrollWidth === clientWidth` in every case).

Also pinned statically by `src/pages/Agency/Edit/styles.test.ts`, which asserts the 800px width, the max-width cap, and the deliberate **absence** of an item min-width.

Full measurement log: `/Users/kio/ORISO/_e2e-artifacts/jobs-2026-08-19/screenshots/job1-legal-editor-width-MEASUREMENT.txt`

kiodreambau:

PR #864, opened 2026-08-20
feat(legal-editor): consent template chooser in the agency DPP function bar

kiodreambau:
## Problem

The department data-protection editor's lower function bar carried three controls: language, topic, version. The consent sentence had no template chooser there, even though the consent sentence is a **field of that policy** (ADR-021 decision 4) — so its chooser belongs in the same bar. The footer now reads **language, consent template, topic, version**.

## Root cause

Not a defect so much as a missing slot: `M3RichTextEditor`'s function bar had no extension point between the language control and `topicSlot`, so the consent module drew its own chooser elsewhere in the form.

## Fix

- **`M3RichTextEditor`**: new optional `consentSlot`, rendered between the language control and `topicSlot`, and added to the bar's render guard so it can hold the bar open on its own. Purely additive — hosts that pass nothing get exactly the bar they had.
- **`DepartmentDataProtectionCard`** fills the slot with the **existing `TemplateSplitButton`** (no second split button was built) and owns the picked template; applying one replaces only the active language's sentence.
- The chooser is **visible but disabled** on read-only surfaces and while an archived version is on screen, per the admin "disable, never hide" rule.
- `useConsentTemplates` holds the template list once, so the two surfaces that can show the chooser cannot drift apart.
- `hideTemplateChooser` keeps the consent module from drawing a second, identical chooser now that the card lifted it into the function bar.

### Scope decision carried over from the commit

The **tenant/platform level was deliberately left out at the owner's request (2026-08-19)**. `LegalText.test.tsx` asserts it passes no `consentSlot`, so nobody "completes" this later by mistake. If you think that test is wrong, that is an owner decision, not a review nit.

## Tests

- All five `M3RichTextEditor` guard suites plus the new `consentSlot` suite: **59 tests**.
- `LegalSettings` and `PlaceholderTemplate` suites: **168 tests**.
- `eslint --max-warnings=0` and `prettier` on the touched files.
- **axe** (wcag2a / 2aa / 21a / 21aa / 22aa) on both new stories at 390x844: **0 violations**.

## Reviewer test plan

- [ ] Storybook → `M3RichTextEditor` → the new consent-slot stories: the function bar reads language, consent template, topic, version in that order.
- [ ] Open a Fachbereich data-protection card; pick a consent template from the split button. Only the **active language's** sentence changes; the others are untouched.
- [ ] Switch the card to read-only: the chooser is still **visible but disabled**, not hidden.
- [ ] Open an **archived version**: same — visible, disabled.
- [ ] Confirm only **one** template chooser is on screen (the consent module must not draw a second one).
- [ ] Open the **tenant/platform** legal text: there is deliberately **no** consent chooser in its function bar.
- [ ] **Mobile 390x844**: the four controls wrap sanely and the bar stays operable.

## Summary by CodeRabbit

* **New Features**
  * Added consent-template selection to department legal text editing.
  * Templates can be applied to the active language without immediately saving changes.
  * Added support for hiding the template chooser when needed.
  * Added read-only and archived-state behavior that disables template selection.

* **Tests**
  * Added coverage for consent controls, template selection, visibility, ordering, and read-only behavior.

### DIFF
--- a/src/components/FormPluginEditor/M3RichTextEditor.consentSlot.test.tsx
+++ b/src/components/FormPluginEditor/M3RichTextEditor.consentSlot.test.tsx
@@ -1,0 +1,94 @@
+import { beforeAll, describe, expect, it, vi } from 'vitest';
+import { render, screen, waitFor } from '@testing-library/react';
+import { M3RichTextEditor } from './M3RichTextEditor';
+
+vi.mock('react-i18next', () => ({
+    useTranslation: () => ({
+        t: (key: string, fallbackOrOptions?: unknown) => {
+            if (typeof fallbackOrOptions === 'string') return fallbackOrOptions;
+            return key;
+        },
+    }),
+}));
+
+beforeAll(() => {
+    Object.defineProperty(window, 'matchMedia', {
+        writable: true,
+        value: vi.fn().mockImplementation((query: string) => ({
+            matches: false,
+            media: query,
+            onchange: null,
+            addEventListener: vi.fn(),
+            removeEventListener: vi.fn(),
+            addListener: vi.fn(),
+            removeListener: vi.fn(),
+            dispatchEvent: vi.fn(),
+        })),
+    });
+    Element.prototype.scrollIntoView = vi.fn();
+});
+
+/**
+ * The lower function bar carries up to four split buttons (Figma 1261-48667 plus
+ * the consent template chooser of ADR-021 decision 4). `consentSlot` is the third
+ * of those to be filled from outside, and — like `topicSlot` — must be able to
+ * hold the bar open on its own.
+ */
+describe('M3RichTextEditor consentSlot', () => {
+    it('renders the consent slot in the lower function bar', async () => {
+        render(<M3RichTextEditor title="Datenschutz" consentSlot={<button type="button">Vorlage 2</button>} />);
+
+        await waitFor(() => {
+            expect(screen.getByRole('button', { name: 'Vorlage 2' })).toBeInTheDocument();
+        });
+    });
+
+    it('orders the bar language → consent → topic → version', async () => {
+        render(
+            <M3RichTextEditor
+                title="Datenschutz"
+                languageSlot={<span data-testid="slot-language">DE</span>}
+                consentSlot={<span data-testid="slot-consent">Vorlage 2</span>}
+                topicSlot={<span data-testid="slot-topic">Sucht</span>}
+                versions={[{ id: 'v1', label: '1. Jul 2026', content: '<p>alt</p>' }]}
+            />,
+        );
+
+        const bar = await screen.findByTestId('m3-editor-function-bar');
+        const language = screen.getByTestId('slot-language');
+        const consent = screen.getByTestId('slot-consent');
+        const topic = screen.getByTestId('slot-topic');
+        const version = screen.getByTitle('legal.m3Editor.versionHistory');
+
+        // All four live in the same bar …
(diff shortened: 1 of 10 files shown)


kiodreambau:
## Visual evidence

**Screenshot:** `/Users/kio/ORISO/_e2e-artifacts/jobs-2026-08-19/screenshots/job2-consent-template-slot-desktop.png`
Storybook `Organisms/M3 Rich Text Editor` → `AgencyFooterReadOnly`, Chromium 1440x900, Node 22.12.0.

It shows the agency function bar with the **consent template chooser ("Vorlage 2", document glyph + split dropdown) sitting between the editor body and the topic chooser**, ahead of `Sucht` (topic) and `Latest Version` (version) — the new ordering this PR introduces.

Two things a reviewer should know about *which* story this is:

- It is the **read-only** story, so the chooser is rendered **greyed and inert**. That is deliberate and is itself one of the PR's claims (the admin "disable, never hide" rule) — it is not a broken control.
- The **enabled** story (`AgencyFooterWithConsentTemplate`) is deliberately framed at **390x844** to prove the bar scrolls without a scrollbar, so at desktop size it renders in a 390px box with the bar scrolled to its right-hand end and the chooser off-frame. Its `play` function is what asserts the full left-to-right order `Sprache → Vorlagenmenü → Fachbereich → Versionsverlauf`. The read-only story was chosen for the still because it is the one that renders the bar at full width.

The language control is absent from this particular story only because it declares a single language.

kiodreambau:

Shirloin:
@kiodreambau 
Two CodeRabbit findings still need to be addressed before merge:

1. Template inserts UI-language text instead of legal-content language — the selected template variant should use the editor's content language, not the UI locale.

2. Selected state sticks to wrong language after department switch — selection should reset when department or language changes.

Please fix both before re-requesting review.

kiodreambau:
The owner re-raised this as **JOB2** on 2026-08-24, so this PR is the one piece of that review
round still outstanding. Verified against `origin/pre-dev` today:

```
grep -c 'consentSlot\|templateSlot' origin/pre-dev:src/components/FormPluginEditor/M3RichTextEditor.tsx
0
```

Everything else from the round (#863 width, #866 sync, #868 scrollbars, #872 head icon) is
merged.

Two notes for whoever finishes this:

1. **The invisible-scrollbar half is already done** — `.functionBar` in
   `M3RichTextEditor.module.scss` carries `flex-wrap:nowrap`, `overflow-x:auto`,
   `scrollbar-width:none`, `&::-webkit-scrollbar{display:none}` and `> * { flex-shrink:0 }`.
   Nothing to add there; if the bar does not scroll in practice the cause is an ancestor
   clipping it.
2. **Reuse `PlaceholderTemplate/TemplateSplitButton.tsx`** rather than building a second split
   button — it is already the M3 split button with this geometry, already honours `disabled`,
   and already has a test. Read-only surfaces keep the control visible-but-inert, per the ORISO
   rule that a hidden control also hides what the level offers.

Target order in the function bar, from the owner's annotation:
`language · consent template · topic · version` — so the new slot renders **after**
`languageControl` and **before** `topicSlot`, and has to join the render guard on
`M3RichTextEditor.tsx:1014` or the bar vanishes when it is the only control present.

**Before merging, note the deployed pre-dev Storybook is stale** (pod from 2026-08-16). An
after-shot taken there today would not show this work even once merged — it needs a rollout
restart first. Detail in ORISO-Admin#874.

shanzaeimran (approved):

PR #866, opened 2026-08-20
fix(legal): keep the consent sentence in sync with the policy it belongs to

kiodreambau:
## Problem

The consent sentence is a field of the data-protection policy (ADR-021 decision 4), so it must publish **with** the policy and never drift from it. Two places let it drift.

### 1. A forking Fachbereich shipped a policy whose consent sentence stayed on the level above

A Fachbereich that had not yet forked from the inherited text was seeded with the inherited **body** but with its **own (empty) consent map**. Publishing forked the policy for good (ADR-014 amendment) and left its consent sentence behind on the level above. This is a real bug found while doing the work, not a theoretical one.

### 2. The blocked-publish notice only appeared after you had already failed

The "cannot publish, `{{legal_links}}` is missing" notice only appeared **after a failed Publish click**. The rule arrived after texts were already live, so a stored sentence can be blocking **on open** — and in a language other than the one on screen, where nothing pointed at it.

## Root cause

- The fork seeded body and consent from **two different sources**.
- The notice was derived from a **failed click** rather than from validation state.

## Fix

- The fork now seeds **both from the same source** and says which level the sentence came from. A Fachbereich that has **already** forked keeps its stored sentence, **blank included** — because blank means the level above still governs at runtime (ADR-021 decision 1).
- The notice is now **derived from the validation state**, and names the blocked languages **the way the admin reads them** instead of as wire codes.

No i18n keys added; the existing `publishBlocked` strings already carry the language list. The client-side check stays a **mirror of the server rule**, so the admin is told before the request rather than by a 400.

## Tests

New/extended coverage for:

- the inline missing-token error,
- the **blank-is-inherited** semantics,
- a **consent-only edit publishing a byte-identical body**,
- `consentText` actually reaching the publish endpoint.

Files: `AgencyLegalTextContainer.consent.test.tsx` (+75), `DepartmentDataProtectionCard.consent.test.tsx` (+150), `usePublishDepartmentDpp.hook.test.tsx` (+22).

## Screenshot / evidence note

This one is **deliberately not screenshotted**. Both halves are behavioural: the fork-seeding bug is invisible until you publish and then inspect *which level* the stored sentence landed on, and the notice change is a change in **when** an already-existing notice appears, not in how it looks. A still image of the notice would prove nothing that the tests do not already prove, and would suggest visual work that was not done. The assertions above are the evidence.

## Reviewer test plan

- [ ] Take a Fachbereich that has **not** forked. Edit and publish. Confirm the published policy carries the consent sentence it inherited — it must **not** stay behind on the level above.
- [ ] Take a Fachbereich that **has** forked with a **blank** consent sentence. Confirm the blank is preserved (blank = the level above governs at runtime) and is not "helpfully" refilled.
- [ ] Store a consent sentence missing `{{legal_links}}` in a **non-active** language, then reopen the editor. The blocked notice appears **immediately on open**, without clicking Publish.
- [ ] Confirm the notice names the language in words (e.g. "German"), **not** a wire code like `de`.
- [ ] Make a **consent-only** edit and publish; the body should go up byte-identical.
- [ ] Confirm you get the client-side block **before** the request, not a server 400.
- [ ] **Mobile 390x844**: the notice with a multi-language list wraps rather than truncating.


kiodreambau:
## Evidence note

No screenshot for this PR — the reasoning is in the **Screenshot / evidence note** section of the description above, and it has not changed after attempting capture:

- The **fork-seeding bug** is invisible until you publish and then inspect *which level* the stored consent sentence landed on. There is no frame in which the bug and the fix look different.
- The **notice change** alters *when* an already-existing notice appears (on open, derived from validation state, rather than after a failed Publish click), not how it looks — plus it now names languages in words rather than wire codes, which is a string change inside the same component.

A still of the notice would look the same before and after and would suggest visual work that was not done. The assertions listed in the description are the evidence.

kiodreambau:

PR #867, opened 2026-08-20
fix(dpa): make the unsigned-contract dialog a real gate and verify on unlock

kiodreambau:
## Problem

The "tenant must sign the contract before doing anything" gate **was not a gate**. The forwarded-pending dialog was a notice:

- it rendered **the whole admin app behind it**,
- it could be dismissed **four independent ways** — "Später", the X, Escape, and a mask click,
- and its own copy promised that *"everything else is already available"*, which contradicted the entire point of the screen.

## Root cause

The dialog was an **overlay over the admin routes** rather than a replacement for them, and the status answer it gated on could be **served from cache** — so even the checks it did make could be stale.

### JOB7 — the gate is a gate

- The waiting dialog **replaces the admin routes** instead of overlaying them, so there is nothing behind it to click. No X, no mask dismiss, no Escape.
- **"Später" is gone**; the only exit is **"Abmelden" / "Log out"**.
- The note now says the platform stays **fully locked** until the signature lands — correcting the contradictory copy.
- The status answer is **never served from cache**: `staleTime`/`gcTime` 0, an explicit cache drop on the first render of a gate (`refetchOnMount` alone still hands render #1 the old value), refetch on focus/reconnect, and a **30s poll** that stops at `VALID` and pauses in a background tab.

### JOB8/JOB9 — the signature can land while the tenant waits

- A `VALID` status arriving after the waiting dialog was shown yields a new **`unlock-confirm`** decision with an **"Unlock platform"** gate, instead of silently dropping the tenant into the admin area.
- That button **re-asks the backend and reads THAT answer**: `VALID` opens the app; anything else returns the tenant to the waiting gate **with the reason**; a failed request falls back to the **fail-closed** `STATUS_UNAVAILABLE` blocker.
- **JOB9 is the acceptance criterion for this button, not separate work** — that is why these three jobs are one PR.

Read-only throughout (**ADR-022**): the gate only `GET`s the DPA status.

## Tests

Verification on **node 22.23.1**:

- vitest `DpaBlocker` + `DpaForwardDialog` + `Modal` + `dpaBlockerGate` + `dpaForward` + `OrganisationDpaStep`: **129 passed**
- **axe-core** wcag2a/2aa/21a/21aa/22aa on **both** gate dialogs: **0 violations**
- `eslint --max-warnings=0` and `prettier --check` on the touched files
- scoped `tsc` (app + storybook configs) on the touched files

## Reviewer test plan

- [ ] As a tenant with an unsigned contract, log in. The waiting gate is **all there is** — the admin app is not rendered behind it.
- [ ] Try to dismiss it **four ways**: there is no X, the mask does not close it, **Escape** does nothing, and there is no "Später".
- [ ] Confirm the only control is **"Abmelden" / "Log out"**, and that it actually logs out.
- [ ] Read the note text: it must say the platform stays fully locked, **not** that "everything else is already available".
- [ ] With the gate open, sign the contract elsewhere. Within ~30s the gate becomes the **"Unlock platform"** dialog — you are **not** silently dropped into the admin area.
- [ ] Click **Unlock platform** while the backend still says not-valid: you are returned to the waiting gate **with the reason** shown.
- [ ] Kill the backend and click **Unlock platform**: you get the fail-closed `STATUS_UNAVAILABLE` blocker, **not** access.
- [ ] Background the tab for a few minutes and return — polling pauses while hidden and resumes on focus; the first render after return is **not** the stale cached status.
- [ ] **Mobile 390x844**: both gate dialogs are readable, the log-out and unlock buttons are reachable, and nothing scrolls behind the gate.


kiodreambau:
## Visual evidence

Storybook `Organisms/DpaBlocker`, Chromium 1440x900, Node 22.12.0. Two screenshots, because this branch has two distinct gates.

**JOB7 — the waiting gate**
`/Users/kio/ORISO/_e2e-artifacts/jobs-2026-08-19/screenshots/job7-9-contract-gate-desktop.png`
Story `PendingSignatureDialog/Default`. Visible in the image:

- **No X, no dismiss affordance** anywhere on the dialog.
- The only actions are **"Log out"** and "Send e-mail" — **"Später" is gone**.
- The corrected note reads *"Until the signature arrives, the platform stays fully locked for your organisation."* — replacing the old copy that promised everything else was already available.
- Nothing is rendered behind the dialog.

**JOB8/JOB9 — the unlock gate**
`/Users/kio/ORISO/_e2e-artifacts/jobs-2026-08-19/screenshots/job7-9-unlock-platform-desktop.png`
Story `UnlockDialog/Default`. Visible in the image:

- "Agreement signed" with an **"Unlock platform"** button — the tenant is *not* silently dropped into the admin area.
- The copy states the re-verification contract explicitly: *"Unlocking verifies the signature against the server once more. The platform only opens when that check confirms the agreement has been signed."*

The cache/poll behaviour and the fail-closed `STATUS_UNAVAILABLE` path are behavioural and are covered by the 129 passing vitest assertions, not by these stills.

kiodreambau:

shanzaeimran (approved):

PR #868, opened 2026-08-20
fix(a11y): take the grey scroll track off the reader and signup surfaces

kiodreambau:
## Problem

The contract document box, the signup wizard and the public page shell all showed a **permanent grey scroll channel** down their right edge on **Firefox and Edge**. macOS paints overlay scrollbars that fade when idle, which is why the report named those two browsers specifically — and why the **signup module**, whose page shell is a full-viewport scroller, looked worst.

## Root cause

The **track**, not the thumb, was the defect: these surfaces had no `scrollbar-width`/track treatment, so browsers that paint classic (non-overlay) scrollbars drew a full-height grey channel whether or not the user was scrolling.

## Fix

- Add a **`quiet` mixin** (`src/styles/_scrollbars.scss`): `scrollbar-width: thin` plus a **transparent track**. It is the idiom the admin sidebar and the mobile editor toolbar already use — now **named and shared** rather than copy-pasted.
- Apply it to the legal reader's viewport (one rule covering the deck card, the fullscreen dialog and the bounded fluid read card), the onboarding sheet body, the DPA blocker overlay and card body, and the forward dialog body.
- `publicLayout.less` repeats the declarations **inline** — LESS cannot use an SCSS partial.

### Two judgement calls carried over from the commit

1. **The thumb is kept, deliberately.** This is **not** the repo's full-hide pattern. These are **vertical readers**: the thumb is the only signal that a contract continues past the fold, and the chapter bar below navigates *between* chapters, not *within* one. The full-hide pattern stays on the **horizontal** control rows, which carry their own edge-fade and arrow affordances.
2. **`scrollbar-gutter: stable` is kept where it was.** It prevents a sideways jump as a step grows, and only the track was ever the defect.

No webkit pseudo-element: **Chromium ignores it once `scrollbar-width` is set**, as already documented in `globalSearchBar.module.scss`.

## Tests

Node 22.12.0:

- vitest on the new `quietScrollbar.styles.test.ts` plus the existing `legalReader` / `DepartmentDataProtectionCard` style tests: **23 passed**
- `DpaBlocker`, `DpaForwardDialog` and `TenantOnboarding` suites: **122 passed**
- reader-related `FormPluginEditor` suites: **44 passed**
- `prettier` and `eslint` clean

## Reviewer test plan

- [ ] **In Firefox or Edge** (this is the point — macOS Chrome/Safari overlay scrollbars hide the defect): open the contract document box. The grey channel is gone; a thin thumb remains.
- [ ] Same in the **signup wizard** — this surface looked worst before.
- [ ] Same on the **public page shell**, the **DPA blocker** and the **forward dialog** body.
- [ ] Confirm the **thumb is still visible** on the legal reader. It must not be fully hidden — it is the only cue the contract continues.
- [ ] Grow a signup step (e.g. trigger a validation error) and confirm there is still **no sideways jump** — `scrollbar-gutter: stable` was kept for this.
- [ ] Check the **horizontal** control rows still use the full-hide pattern; this PR must not have changed them.
- [ ] **Mobile 390x844**: reader and signup scroll normally, no stray track.

### DIFF
--- a/src/components/DpaBlocker/styles.module.scss
+++ b/src/components/DpaBlocker/styles.module.scss
@@ -1,4 +1,5 @@
 @use '../../styles/breakpoints' as bp;
+@use '../../styles/scrollbars' as scrollbars;
 
 /* Global DPA blocker (TEN-INV-U10, #572).
  *
@@ -30,6 +31,12 @@
     background: var(--m3-surface-container-low, #f6f3f3);
     inset: 0;
     overflow-y: auto;
+
+    /* JOB10: the blocker covers the whole viewport, so its classic scrollbar
+       read as the window's own — a full-height grey track beside the contract
+       the admin is being asked to sign. */
+
+    @include scrollbars.quiet;
 }
 
 .card {
@@ -76,7 +83,12 @@
         padding: 0 4px;
         margin: 0 -4px;
         overflow-y: auto;
+
+        /* Gutter stays so the sign form does not jump sideways as it grows;
+           JOB10 only takes the track away. */
         scrollbar-gutter: stable;
+
+        @include scrollbars.quiet;
     }
 }
 
--- a/src/components/DpaForwardDialog/styles.module.scss
+++ b/src/components/DpaForwardDialog/styles.module.scss
@@ -1,3 +1,5 @@
+@use '../../styles/scrollbars' as scrollbars;
+
 /* House rule (owner, 2026-08-18): a dialog never touches the top or bottom edge
    of the viewport. The sheet is therefore bounded and its BODY is the scroller,
    so the hero icon, the title and the action row stay put while only the content
@@ -22,6 +24,8 @@
             /* A long preview must not chain its scroll to the page behind the mask. */
             overscroll-behavior: contain;
             scrollbar-gutter: stable;
+
+            @include scrollbars.quiet;
         }
     }
 }
--- a/src/components/FormPluginEditor/M3RichTextEditor.module.scss
+++ b/src/components/FormPluginEditor/M3RichTextEditor.module.scss
@@ -2,6 +2,8 @@
 // node 1:34903). Mirrors the M3 sys tokens; brand primary #273270, M3 seed
 // primary #a5000a (publish accent).
 
+@use '../../styles/scrollbars' as scrollbars;
+
 $m3-surface: #eae7e8;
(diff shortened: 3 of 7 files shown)


kiodreambau:
## Evidence — measured, not photographed

**There is no screenshot for this PR, and here is why: the defect cannot be photographed on macOS.**

The story `organisms-dpablocker--desktop-long-dpa-text-with-chapters` (the "Contract documents" box from the owner's report) was rendered and the two scrolling surfaces measured live, with and without the two declarations this PR adds:

| state | element | gutter | `scrollbar-width` | `scrollbar-color` |
|---|---|---|---|---|
| declarations reverted (simulated *before*) | `_cardBody_` | **0 px** | `auto` | `auto` |
| declarations reverted (simulated *before*) | `_editorContentScroll_` | **0 px** | `auto` | `auto` |
| this branch (*after*) | `_cardBody_` | **0 px** | `thin` | `rgb(196,199,200) rgba(0,0,0,0)` |
| this branch (*after*) | `_editorContentScroll_` | **0 px** | `thin` | `rgb(196,199,200) rgba(0,0,0,0)` |

Two conclusions:

1. **The fix reaches the intended elements.** Both scrolling surfaces resolve to `scrollbar-width: thin` with a **fully transparent track** (alpha 0) — exactly what the `quiet` mixin declares, and it lands on the reader viewport *and* the card body.
2. **The defect is invisible on this machine.** The gutter is **0 px in both states**: macOS paints overlay scrollbars that occupy no layout width and are not composited into a screenshot. That is precisely why the original report named **Firefox and Edge** — on Windows/Linux the scrollbar is classic and the grey track is always on. Forcing Firefox with `ui.useOverlayScrollbars=0` was also tried and still produced no painted scrollbar in a headless macOS capture.

Any "before/after scrollbar" image produced here would have been **two identical pictures with no scrollbar in either** — it would prove nothing, so none was attached rather than passing off a meaningless image as evidence.

**Please review this one in Firefox or Edge on Windows/Linux**, where the change is actually visible.

Full measurement log: `/Users/kio/ORISO/_e2e-artifacts/jobs-2026-08-19/screenshots/job10-quiet-scrollbars-MEASUREMENT.txt`

kiodreambau:

shanzaeimran (approved):

PR #869, opened 2026-08-20
feat(dpa-forward): frame the forwarded mail as itself and pin its footer in Storybook

kiodreambau:
## Problem

The forward dialog previewed its mail **without saying which mail it is**. With no `kind`, `InviteEmailPreviewService` falls back to `TENANT_INVITE`, so the sample call-to-action rendered against the **admin console** — while a DPA signer belongs on the **app host**.

## Root cause

The preview request simply omitted `kind`. The house frame — and with it the footer (brand name, Impressum / Datenschutz, automated-send note) — is applied by the **backend renderer** for every kind; it is not composed here, **and must not be**.

## Second, pre-existing bug found while doing this

The story file **declared no MSW handler** for the preview endpoint. Without one, the request bypassed to the Storybook origin and **every preview-bearing story silently rendered the error state** instead of a mail. That was not caused by this work — it was already true, and nothing failed loudly enough to notice.

## Fix

- Pass **`kind="DPA_FORWARD"`** to the mail preview, with a test on the request.
- Add **`ForwardedMailWithFooter`** as a variant story of the existing `Organisms/DpaForwardDialog`, asserting the footer in the rendered mail.
- Give the story file an **MSW handler** for the preview endpoint (fixes the bug above).
- Add **`brandedEmailStoryPreview`**: substitutes only the per-mail slots (title, preheader, subject, content) into a **checked-in verbatim backend response**, so the footer shown in Storybook is the **real** one rather than markup authored in this repo.
- Correct the `DpaLegalReader` story doc, which still described the sticky chapter bar **reversed** by the 2026-08-19 owner demo.

## Tests

Node 22.12.0:

- vitest **310 passed / 32 files** across `DpaForwardDialog`, `DpaLegalForm`, `DpaBlocker`, `EmailPreview`, `FormPluginEditor` and `TenantOnboarding`
- `eslint` and `prettier` clean on the changed files

## Reviewer test plan

- [ ] Storybook → `Organisms/DpaForwardDialog` → **`ForwardedMailWithFooter`**: the preview shows a real mail with the branded footer (brand name, Impressum / Datenschutz, automated-send note).
- [ ] Confirm the other preview-bearing stories in that file now render a **mail** and not the **error state** — this is the pre-existing MSW bug.
- [ ] Inspect the outgoing preview request: it carries `kind=DPA_FORWARD`.
- [ ] Confirm the sample call-to-action now points at the **app host**, not the admin console.
- [ ] Sanity-check that the footer markup comes from the **checked-in backend fixture** and is not hand-authored in this repo — that is the whole point of `brandedEmailStoryPreview`.
- [ ] Read the corrected `DpaLegalReader` story doc against the actual sticky chapter bar behaviour.
- [ ] **Mobile 390x844**: the mail preview scales and the footer does not overflow the dialog.

### DIFF
--- a/src/components/DpaForwardDialog/DpaForwardDialog.stories.tsx
+++ b/src/components/DpaForwardDialog/DpaForwardDialog.stories.tsx
@@ -2,8 +2,13 @@ import type { Meta, StoryObj } from '@storybook/react-vite';
 // eslint-disable-next-line import/no-unresolved -- SB10 subpath export, invisible to the eslint import resolver
 import { expect, userEvent, waitFor, within } from 'storybook/test';
 import { ThemeProvider } from '@mui/material/styles';
+import { http, HttpResponse } from 'msw';
 import { orisoMuiTheme } from '../../theme/orisoMuiTheme';
 import { PHONE_390 } from '../DpaLegalForm/dpaStoryText';
+import {
+    INVITE_EMAIL_PREVIEW_ENDPOINT,
+    renderBrandedEmailStoryPreview,
+} from '../EmailPreview/brandedEmailStoryPreview';
 import { DpaForwardDialog } from './DpaForwardDialog';
 import { DpaForwardError, DpaForwardLink, DpaForwardOutcome } from '../../api/tenantOnboarding/dpaForward';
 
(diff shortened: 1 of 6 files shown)


kiodreambau:
## Visual evidence

**Screenshot:** `/Users/kio/ORISO/_e2e-artifacts/jobs-2026-08-19/screenshots/job11-forwarded-mail-footer-desktop.png`
Storybook `Organisms/DpaForwardDialog` → `ForwardedMailWithFooter`, Chromium 1440x900 (tall capture, preview scrolled to its end), Node 22.12.0.

The image proves the three claims of this PR:

1. **The preview renders a mail, not the error state.** This is the pre-existing MSW-handler bug — before this branch, every preview-bearing story in this file silently rendered an error.
2. **The branded footer is present and is the real backend frame:** `ORISO` / `Impressum · Datenschutz` / *"Diese E-Mail wurde automatisch versendet. Bitte antworten Sie nicht darauf."*
3. **`kind="DPA_FORWARD"` is taking effect** — the signing link in the body points at the **app host** (`https://app.oriso-dev.site/dpa-sign/…`), not the admin console, which is what the old `TENANT_INVITE` fallback produced.

### One observation for the reviewer

Inside the fixture-rendered mail the call-to-action button still reads **"Einladung annehmen"** and points at `https://app.oriso.org/account-invite/SAMPLE-PREVIEW-TOKEN`. That is consistent with how `brandedEmailStoryPreview` is documented to work — it substitutes only title, preheader, subject and content into a verbatim checked-in backend response, so the CTA slot is whatever the captured fixture had. Worth a sanity check that this is intended for the story, since a reader glancing at the preview sees an invite CTA above a DPA signing link.

kiodreambau:

shanzaeimran (approved):

PR #870, opened 2026-08-20
fix(legal): stop "Save draft" from publishing the agency-wide legal text

kiodreambau:
## Problem

**This is the most serious of the batch.** On the **"Alle Fachbereiche"** selection, the editor's **"Entwurf speichern"** ("Save draft") button **published the live legal text**.

No confirmation, no version note, and the exact opposite of what the label promises. On a legal-text surface that is the worst kind of bug: **it succeeds quietly.**

## Root cause

The draft action reached `onSaveAgencyWide`, which writes the agency record — and the **`publish` argument was dropped on the floor**. So "save a draft" and "publish this to the world" took the same path.

## Fix

Give the agency-wide text the **same device-local draft the Träger level already has**. That is the house answer to this exact situation: the tenant record has no backend draft state either, which is why `useLegalDraft` exists — and its own comment says so.

- `publish === false` now parks the wording in local storage and **leaves the record alone**.
- `publish === true` writes the record **and discards the parked copy**, so a stale draft cannot re-seed the editor afterwards.
- A parked draft **wins over the stored text** when seeding.

### Why not the alternatives (carried over from the commit)

Both alternatives were worse. **Hiding the action** removes a real capability and makes this level the odd one out. **Confirming before the write** keeps a mislabelled button and merely apologises for it.

### The collision this avoids

The draft scope carries the agency id — `<tenantId>:<userId>:agency:<agencyId>`. Both levels store a `privacy` document for the same user, so **without the agency id a Beratungsstelle's parked wording and the Träger's would share one key and silently overwrite each other**.

## Tests

- **4 files / 31 passing.**
- **Reverting only the production file fails 5 of the 6 new assertions** — the tests genuinely pin the fix.
- Consumers (`DepartmentDataProtectionCard`, `LegalText`): **5 files / 62 passing**.
- The three existing container suites gained a `useUserData` mock: the container now reads the opaque user id, and those suites mock every other hook, so without it the real react-query hook runs and the render dies on *"No QueryClient set"*.

### Note for future test authors (from the commit — worth keeping)

**jsdom's `Storage` does not enumerate through `Object.keys`** — it returns `[]` while `length` is `1`. The draft helper reads it through the Storage API instead, or it would report "nothing was saved" for a draft that saved perfectly well.

## Reviewer test plan

- [ ] Select **"Alle Fachbereiche"**, edit the legal text, click **"Entwurf speichern"**. Confirm the **live** text is unchanged — check the public-facing text, not just the editor.
- [ ] Reload the page. The parked draft **re-seeds the editor** and wins over the stored text.
- [ ] Now click **publish**. The record is written **and** the parked draft is discarded — reload again and confirm a stale draft does not come back.
- [ ] As the **same user**, park a draft at **Träger** level and a different one on a **Beratungsstelle**, both for `privacy`. Confirm they do **not** overwrite each other (this is what the agency id in the scope is for).
- [ ] Park drafts on **two different agencies** as the same user; confirm they stay separate.
- [ ] Confirm the Träger level behaves exactly as before — this reuses `useLegalDraft`, it should not have changed.
- [ ] **Mobile 390x844**: the draft/publish actions are both reachable and still distinctly labelled.

### DIFF
--- a/src/components/Tenants/LegalSettings/components/AgencyLegalTextContainer/AgencyLegalTextContainer.consent.test.tsx
+++ b/src/components/Tenants/LegalSettings/components/AgencyLegalTextContainer/AgencyLegalTextContainer.consent.test.tsx
@@ -30,6 +30,12 @@ vi.mock('../../../../../hooks/useTranslateLegalContent.hook', () => ({
 vi.mock('../../../../../hooks/useUserPermission', () => ({
     useUserPermissions: () => ({ can: () => true, permissions: {} }),
 }));
+vi.mock('../../../../../hooks/useUserData.hook', () => ({
+    // The container reads the opaque user id to scope its device-local draft; without
+    // this mock the real react-query hook runs and the render dies on "No QueryClient".
+    useUserData: () => ({ data: { id: 'user-7' }, isLoading: false }),
+    USER_DATA_KEY: 'user-data',
+}));
 vi.mock('../../../../../hooks/useLegalTextVersions.hook', () => ({
     useLegalTextVersions: () => ({ data: [], isError: false }),
 }));
--- a/src/components/Tenants/LegalSettings/components/AgencyLegalTextContainer/AgencyLegalTextContainer.draft.test.tsx
+++ b/src/components/Tenants/LegalSettings/components/AgencyLegalTextContainer/AgencyLegalTextContainer.draft.test.tsx
@@ -1,0 +1,165 @@
+import { render, screen, waitFor } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const h = vi.hoisted(() => ({
+    card: vi.fn(),
+    saveAgencyWide: vi.fn(),
+}));
+
+vi.mock('react-i18next', () => ({
+    useTranslation: () => ({ t: (key: string) => key, i18n: { language: 'de' } }),
+}));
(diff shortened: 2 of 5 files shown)


kiodreambau:
## Evidence note — why there is no screenshot

**This change cannot be shown in a still image, and the reason is the nature of the bug itself: the fix is a button that *stops* doing something.**

Before, "Entwurf speichern" published the live legal text. After, it parks a device-local draft and leaves the record alone. Both states look **identical** on screen at the moment of the click — the difference is whether the public-facing legal text changed afterwards. A screenshot of the editor proves nothing either way.

There is also no Storybook story for `AgencyLegalTextContainer` (confirmed: the directory contains only `index.tsx`, `styles.module.scss` and four test files), so there is no component-level surface to photograph.

The evidence that does exist is behavioural and is in the tests:

- **4 files / 31 passing** on the new draft behaviour.
- **Reverting only the production file fails 5 of the 6 new assertions** — the tests genuinely pin the fix rather than merely accompanying it.
- Consumers (`DepartmentDataProtectionCard`, `LegalText`): **5 files / 62 passing**.

The one thing worth a reviewer's own hands is the **scope-collision case**, which no test can make obvious: park a draft at Träger level and another on a Beratungsstelle **as the same user, both for `privacy`**, and confirm they no longer overwrite each other. That is what the agency id in `<tenantId>:<userId>:agency:<agencyId>` is for.

kiodreambau:

shanzaeimran (approved):

PR #872, opened 2026-08-21
fix(legal): full-width consent description, house icons, small split button

kiodreambau:
## Problem

Four owner findings on the legal / consent surfaces, reported 2026-08-20 with screenshots.

## What was wrong, and what was not

**The background colour was not a defect.** The owner reported an "extremely ugly" background and suspected a broken token. Measured on the live story at 1440x900, the dialog sheet is `rgb(234, 231, 232)` = `#eae7e8`, exactly the token he named. `applyAdminTheme` does not run in Storybook at all, and where it does run it writes that literal, so it cannot drift. **No change was made** — repainting a correct surface would have been worse than the complaint.

Worth a separate decision, not fixed here: the two inset panels are `#fcf9f9`, the same white-sticker-on-grey pattern rejected once before for the search pill (`app.css:200-215`). They are left as-is because the preview panel deliberately simulates the registration form; recolouring it would make the preview misrepresent the real thing.

**The other three were real:**

- **Description too narrow.** `Modal` caps `.description` at `52ch`, which is right for confirm dialogs. Added an opt-in `descriptionFullWidth`, set only on `PlaceholderTemplateDialog`. Lifting the cap alone was not enough — `.titleBlock` is a centred flex column, so it also needed `align-self: stretch`. **457px → 1032px, 2 lines → 1.** Non-opted-in dialogs verified unchanged at 459.27px.
- **Wrong icons.** `DepartmentSelect` used MUI `Groups`, a *people* glyph, for a control that picks a *theme*. `TopicIcon` already existed in `CustomIcons/LegalIcons.tsx`, backed by an asset byte-identical to the master file — it simply was not used. Checkbox icons did not exist and were added from the master set.
- **Split button too large.** `TemplateSplitButton` used SplitButton's `medium` default (56px) beside a 36px version control. Now `small` (40px).

## Icon naming

Added as `check-box.svg` / `check-box-filled.svg`, not the master's `check_box_checked_*`. The catalogue generator's `normalizeFamily` strips `filled` but not `checked`, so the master naming would have split the pair into two unrelated families instead of pairing them like `topic`. Outline variant everywhere: the repo's rule is filled = *selected*, and every existing Modal hero icon is outlined.

## Catalogue drift

Committed as generated, as the gate requires. It also credits `LegalIcons.tsx` to `ui-icon:x:base` — a **false positive**: `usageFiles` uses a bare `content.includes(fileName)` and `x.svg` is a substring of `check-bo**x.svg**`. Pre-existing, not introduced here (pre-dev already credits `x.svg` to 10 files when one imports it). Filed separately.

## Tests

204 tests / 19 files pass on Node 22.12.0. ESLint, Prettier, Stylelint clean. **axe: 0 violations across all 16 touched stories.**

Two pre-existing axe violations on `DepartmentDataProtectionCard`'s consent story (antd `.ant-tag` contrast, unlabelled `role="img"`) were proven pre-existing by stashing and re-running against a clean tree, and filed separately.

## Reviewer test plan

- [ ] Consent dialog description spans the full box width at 1440 / 820 / 390
- [ ] Other dialogs (e.g. `organisms-dpaforwarddialog--default`) still wrap at the 52ch cap
- [ ] Consent dialog head and split button both show the checkbox icon
- [ ] The theme picker shows the topic icon, not a people glyph
- [ ] Split button sits at the same height as the version control beside it
- [ ] No axe violations on the touched stories

Refs the owner's 2026-08-20 review round.


kiodreambau:

shanzaeimran (approved):

